PR #5, opened 2016-02-06
Setting setsid for the plugin process

diptanu:
This isolates the plugin process from the process group of the plugin host and hence prevents the plugin and the sub-processes it launches to recieve interrupts that the plugin hosts receive. 

Although the plugin ignores any interrupts that it receives but it's sub-processes might not be designed to ignore interrupts.

For ex - When we launched a redis process using the plugin in Nomad, an interrupt to the Nomad client interrupted the redis process and the redis process died along with the Nomad client since it was still in the process group of the plugin host.

### DIFF
--- a/client.go
+++ b/client.go
@@ -335,6 +335,7 @@ func (c *Client) Start() (addr net.Addr, err error) {
 	cmd.Stdin = os.Stdin
 	cmd.Stderr = stderr_w
 	cmd.Stdout = stdout_w
+	c.isolateCmd(cmd)
 
 	log.Printf("[DEBUG] Starting plugin: %s %#v", cmd.Path, cmd.Args)
 	err = cmd.Start()
--- a/cmd_isolate_unix.go
+++ b/cmd_isolate_unix.go
@@ -1,0 +1,14 @@
+package plugin
+
+import (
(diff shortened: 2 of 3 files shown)


diptanu:
@mitchellh Not sure if this is needed for the other projects which uses the plugin. We could perhaps make this a plugin configuration if this is only needed for Nomad's use case.

mitchellh:
I thought about including this originally but unfortunately we can't because some plugins do actually rely on receiving signals and being attached to the parent. :( Maybe eventually.

mitchellh:
This doesn't fix the pid waiting though right?

diptanu:
@mitchellh #6 fixes the pid waiting issue.

So what is your suggestion regarding this? If we don't do this the user processes are vulnerable to getting interrupt signals when nomad client receives them.

I could potentially set the setsid on the user process which the plugin process starts and see how that works out, pretty sure there will be newer edge cases in that case :)

mitchellh:
@diptanu You can do this before aclling` plugin.Serve`, so just still do itin your main() in your plugin.

diptanu:
Oh yeah that would work. Thanks!

PR #6, opened 2016-02-07
Implemented a poll based wait impl for non-child processes

diptanu:
Introduces a Wait utility method which polls a process to check whether it's still alive or not. This is useful when the process which needs to be tested is not a child process of the caller.


mitchellh:
This looks great! I'm going to add some tests, but otherwise perfect.

wait_windows.go
```diff
@@ -0,0 +1,45 @@
+package plugin
+
+import (
+	"os"
+	"syscall"
+	"time"
+)
+
+const STILL_ACTIVE = 259
```
sethvargo: This doesn't appear to match Go's recommendations for constant naming.

mitchellh:
A few updates I made to style:

* Pull the platform-independent stuff into `process.go` and call other platform-specific functions in `process_PLATFORM.go`, which is more idiomatic.

* Pull more into constants for Windows. @sethvargo it isn't idiomatic Go style but it matches the MSDN docs directly for what the returned constant is so we'll go with it.

* Made the constants for Windows unexported, don't want people somehow inadvertently seeing them in docs.

mitchellh:
Merging!

PR #8, opened 2016-02-08
Implement connections over Windows npipes.

calavera:
Leave TCP behavior intact when MIN and MAX ports are set in windows.

https://msdn.microsoft.com/en-us/library/windows/desktop/aa365590

Signed-off-by: David Calavera <david.calavera@gmail.com>

### DIFF
--- a/client_posix.go
+++ b/client_posix.go
@@ -1,0 +1,9 @@
+// +build !windows
+
+package plugin
+
+import "net"
+
+func dialAddress(addr net.Addr) (net.Conn, error) {
+	return net.Dial(addr.Network(), addr.String())
+}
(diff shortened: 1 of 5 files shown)


mitchellh:
Cool! We've been interested in this. I'm sure you have a reason so I'd like to ask: is there a reason this is superior over loopback connections? On Windows in particular (not POSIX), 127.0.0.1 connections are special cased to be extremely performant. I'm wondering if named pipes have some benefit over this.

calavera:
Tbh, I'm not very sure about the advantages. I know that MS prefers to use them over loopback connections because they're more similar to unix pipes:

https://github.com/docker/docker/pull/19911

mitchellh:
I'll do some research and find a Windows machine to verify this on. This otherwise looks good to me.

PR #12, opened 2016-06-02
unix domain sockets should be deleted on listener close

mitchellh:
We were leaving the Unix domain sockets in the temp folders for the system to handle before. On long running systems this was causing a problem. We now clean them up manually on listener close.


dadgar:
LGTM

PR #14, opened 2016-06-07
client should attempt graceful quit

mitchellh:
We were only calling process.Kill before which kills the process without
allowing it to perform any cleanup (defers are never called). This was
leaving around a lot of temporary files over time.

We now add a graceful quit method to the control stream that we attempt
to call first. If that fails, we still force quit, but at least this
gives the process a chance to gracefully exit.

🚧 **NOTE:** Don't merge this. I want to re-review this myself personally and make sure I'm happy with the way we're doing this. I posted this so @alex can review and test with Nomad.

### DIFF
--- a/client.go
+++ b/client.go
@@ -243,8 +243,16 @@ func (c *Client) Kill() {
 		return
 	}
 
-	// Kill the process
-	c.process.Kill()
+	// Close the client to cleanly exit the process
+	client, err := c.Client()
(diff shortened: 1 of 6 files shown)


mitchellh:
Confirmed by @alex to fix the Nomad issue.

mitchellh:
Added a test, merging.

PR #19, opened 2016-11-04
Split the plugin-example into two binaries.

### LINKED ISSUE #11: more examples
Hi,

Would it be possible to see more implementation examples? I'm wrapping my head around on how to create dynamically loaded plugins ;/

skx:
This reflects actual usage, and will close #11.

### DIFF
--- a/examples/basic/README.md
+++ b/examples/basic/README.md
@@ -1,0 +1,15 @@
+Plugin Example
+--------------
+
+Compile this driver via:
+
+    go build .
+
+Compile the plugin itself via:
+
+    cd plugin/ && go build .
+
+You can then launch the plugin sample via:
+
+    ./basic
+
--- a/examples/basic/plugin/plugin.go
+++ b/examples/basic/plugin/plugin.go
@@ -1,46 +1,19 @@
 package main
 
 import (
-	"fmt"
-	"io/ioutil"
-	"log"
 	"net/rpc"
-	"os/exec"
-
 	"github.com/hashicorp/go-plugin"
 )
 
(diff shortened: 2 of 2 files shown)


examples/basic/main.go
```diff
@@ -48,15 +41,6 @@ func main() {
...
-	// so that the host and our plugin can verify they can talk to each other.
-	// Then we set the plugin map to say what plugins we're serving.
-	plugin.Serve(&plugin.ServeConfig{
-		HandshakeConfig: handshakeConfig,
-		Plugins:         pluginMap,
-	})
-}
 
 // handshakeConfigs are used to just do a basic handshake between
 // a plugin and host. If the handshake fails, a user friendly error is shown.
```
xiaoyao1991: Are we supposed to remove the plugin implementations below?

examples/basic/main.go
```diff
@@ -26,7 +19,7 @@ func main() {
 	client := plugin.NewClient(&plugin.ClientConfig{
 		HandshakeConfig: handshakeConfig,
 		Plugins:         pluginMap,
```
xiaoyao1991: I guess this is a more general usage question. Do we really need these configs, especially `pluginMap` here? That means it needs to know what plugins are available at compile time right? Kinda destroy the whole purpose of having plugins

Raggaer: I think the same. is this mandatory?

mitchellh:
Whoops, I saw #20 first (top of list :( ). Sorry. Merged that which should fix the same issue! Thanks!

PR #25, opened 2017-04-12
Plugin Checksumming & TLS support 

briankassouf:
This PR adds support for verifying the integrity of a plugin by matching the binary's checksum with an expected checksum. It introduces the notion of a `SecureConfig` which stores the expected checksum and selected hashing method.

It also adds support for providing a TLS config to both the client and server of a plugin.

### DIFF
--- a/client.go
+++ b/client.go
@@ -355,8 +355,8 @@ func (c *Client) Start() (addr net.Addr, err error) {
 			return nil, fmt.Errorf("Only one of Cmd or Reattach must be set")
 		}
 
-		if secureSet == attachSet {
-			return nil, fmt.Errorf("Only one of Cmd or SecureConfig can be set")
+		if secureSet && attachSet {
+			return nil, fmt.Errorf("Only one of Reattach or SecureConfig can be set")
 		}
 	}
 
--- a/client.go
+++ b/client.go
@@ -38,15 +38,22 @@ var (
 	// ErrProcessNotFound is returned when a client is instantiated to
 	// reattach to an existing process and it isn't found.
 	ErrProcessNotFound = errors.New("Reattachment process not found")
+
 	// ErrChecksumsDoNotMatch is returned when binary's checksum doesn't match
 	// the one provided in the SecureConfig.
 	ErrChecksumsDoNotMatch = errors.New("checksums did not match")
+
 	// ErrSecureNoChecksum is returned when an empty checksum is provided to the
 	// SecureConfig.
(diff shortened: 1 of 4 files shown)


client.go
```diff
@@ -384,6 +449,14 @@ func (c *Client) Start() (addr net.Addr, err error) {
 	cmd.Stderr = stderr_w
 	cmd.Stdout = stdout_w
 
+	if c.config.SecureConfig != nil {
+		if ok, err := c.config.SecureConfig.Check(cmd.Path); err != nil {
+			return nil, err
```
mitchellh: Let's wrap this so that its a little more obvious where this is coming from

client.go
```diff
@@ -35,6 +38,15 @@ var (
 	// ErrProcessNotFound is returned when a client is instantiated to
 	// reattach to an existing process and it isn't found.
 	ErrProcessNotFound = errors.New("Reattachment process not found")
+	// ErrChecksumsDoNotMatch is returned when binary's checksum doesn't match
```
mitchellh: Nitpick: I prefer a newline between the comment key and the next comment:

```go
const (
  // Foo
  Foo

  // Bar
  Bar
)
```

client.go
```diff
@@ -119,6 +138,43 @@ type ReattachConfig struct {
 	Pid  int
 }
 
+// SecureConfig is used to configure a client to verify the integrity of an
+// executable before running. It does this by verifying the checksum is
+// expected. Hash is used to specify the hashing method to use when checksumming
+// the file.  The configuration is verified by the client by calling the
+// SecureConfig.Check() function.
```
mitchellh: I'd recommend adding more color to this comment to explain how to do this securely. Simply setting this won't buy you "security" as you know (since it depends on how this data came in and how the binary was installed and so on) but I think it'd be prudent to be explicit about this here, nicely. :)

client.go
```diff
@@ -318,9 +378,14 @@ func (c *Client) Start() (addr net.Addr, err error) {
...
 		cmdSet := c.config.Cmd != nil
 		attachSet := c.config.Reattach != nil
+		secureSet := c.config.SecureConfig != nil
 		if cmdSet == attachSet {
 			return nil, fmt.Errorf("Only one of Cmd or Reattach must be set")
 		}
+
+		if secureSet && attachSet {
+			return nil, fmt.Errorf("only one of Reattach or SecureConfig can be set")
+		}
```
mitchellh: I may have missed it but is there a test to cover this? Should be pretty simple but would be worth covering since the implication is high.

PR #33, opened 2017-07-22
Support hclog format output on plugins

calvn:


client.go
```diff
@@ -732,9 +734,36 @@ func (c *Client) logStderr(r io.Reader) {
 		line, err := bufR.ReadString('\n')
 		if line != "" {
 			c.config.Stderr.Write([]byte(line))
-
 			line = strings.TrimRightFunc(line, unicode.IsSpace)
-			c.logger.Named(filepath.Base(c.config.Cmd.Path)).Debug(line)
+
+			l := c.logger.Named(filepath.Base(c.config.Cmd.Path))
+			// If output is not JSON format, print directly as error
+			if !isJSON(line) {
```
calvn: I ended up not needing to add anything to the handshake since this allows the output to be printed as-is on the host if hclog format (i.e. JSON formatted output) is not supported on the plugin.

calvn:
@mitchellh  The funcs on `output.log` could use some tests, but I wanted to run this by you first to make sure everything looks right.

client.go
mitchellh: Awesome. This is actually even better because non-Go plugins (that don't have access to go-hclog) can just output structured JSON to still get structured logging on the host. So this is a net win for sure.

client.go
```diff
@@ -732,9 +734,36 @@ func (c *Client) logStderr(r io.Reader) {
...
+				l.Debug("Error from plugin", "error", line)
```
mitchellh: Log lines are not necessarily errors so I'd change this to "log from plugin" "line"

client.go
```diff
@@ -745,3 +774,8 @@ func (c *Client) logStderr(r io.Reader) {
 	// Flag that we've completed logging for others
 	close(c.doneLogging)
 }
+
+func isJSON(str string) bool {
```
mitchellh: It might be better to actually return `*json.RawMessage` (nil meaning not valid JSON) and then just unmarshaling that into the final destination structure to avoid the double-JSON parse.

output.go
```diff
@@ -0,0 +1,102 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"strconv"
+	"time"
+)
+
+// Output is the JSON payload that gets sent to Stderr from the plugin to the host
+type Output struct {
```
mitchellh: Given this is exported (which seems fine), I would name this something more descriptive like `LogEntry`. 

`Output` is just too generic. :)

output.go
```diff
@@ -0,0 +1,102 @@
...
+	Message   string    `json:"message"`
+	Level     string    `json:"level"`
+	Timestamp time.Time `json:"timestamp"`
+	KVPairs   []*KVPair `json:"kv_pairs"`
+}
+
+// KVPair is a key value pair within the Output payload
+type KVPair struct {
```
mitchellh: Same here, would maybe do `LogEntryKV`

server.go
```diff
@@ -101,9 +100,6 @@ func Serve(opts *ServeConfig) {
 		os.Exit(1)
 	}
 
-	// Logging goes to the original stderr
-	log.SetOutput(os.Stderr)
```
mitchellh: I think we still want to do this in case they're not using hclog, right? This shouldn't affect the hclog changes below.

output.go
```diff
@@ -0,0 +1,102 @@
...
+		result = append(result, kv.Value)
+	}
+
+	return result
+}
+
+// Payload returns a payload given a message, level, and kv pairs/
+// If unable to parse KVs, it sets and returns the error message
+// as the payload.
+func payload(message string, level string, kvs ...interface{}) string {
```
mitchellh: If we're exporting the structures above, it may be useful to export this as well so that plugins that don't want to use hclog plugins can still output hclog-valid JSON payloads. If we do that, we should again rename it.

There is an argument though for not exporting ANYTHING log related to encourage Go users to use hclog. That might be worth it?

output.go
```diff
@@ -0,0 +1,102 @@
...
+	"encoding/json"
+	"fmt"
+	"strconv"
+	"time"
+)
+
+// Output is the JSON payload that gets sent to Stderr from the plugin to the host
+type Output struct {
+	Message   string    `json:"message"`
+	Level     string    `json:"level"`
```
jefferai: Do we want to provide preexisting levels that match hclog's levels for ease of use? Even if not technically less typing, simply providing them will give a hint to the user as to which levels are supported in a built-in fashion.

output.go
calvn: I am fine with either approach. We are enforcing hclog to be used on the host side, so it might make sense to enforce hclog on the plugin side as well.

calvn:
@mitchellh I updated the grpc-related tests so they should all be passing now. I need to update the docs to mention hclog support, but the technical implementation should be good to go.

PR #37, opened 2017-07-31
Fix the Python example

sbourlon:
Trying to fix the Python example.

```
% python --version                    
Python 3.6.2
```

Test:
```
cd examples/grpc
go build -o kv
export KV_PLUGIN="python plugin-python/plugin.py"
./kv put hello world
./kv get hello
```
will return:
```
b'world'

Written from plugin-python
```


ethantkoenig:
The `127.0.0.1` change is also necessary for Python 2 (I'm using 2.7.12 on Ubuntu 16.04)

mitchellh:
Thanks! :D

PR #39, opened 2017-08-07
Fix #38: SIGINT gives panic when parsing log output

### LINKED ISSUE #38: SIGINT gives panic when parsing log output
When closing the server with CTRL+c, a panic occurs:
```
panic: interface conversion: interface is float64, not string

goroutine 131 [running]:
panic(0xc524e0, 0xc4209df100)
	/home/bro/programmer/go/go/src/runtime/panic.go:500 +0x1a1
github.com/hashicorp/go-plugin.parseJSON(0xc420445ef0, 0x84, 0x1, 0x1420360, 0xc4209fcbe0)
	/home/bro/programmer/fabriscale/go/ws/src/github.com/hashicorp/go-plugin/log_entry.go:118 +0x4fc
github.com/hashicorp/go-plugin.(*Client).logStderr(0xc4205172d0, 0x1411be0, 0xc420036200)
	/home/bro/programmer/fabriscale/go/ws/src/github.com/hashicorp/go-plugin/client.go:746 +0x277
created by github.com/hashicorp/go-plugin.(*Client).Start
	/home/bro/programmer/fabriscale/go/ws/src/github.com/hashicorp/go-plugin/client.go:544 +0xb3c
```

bendikro:


jefferai:
I think this actually masks the problem rather than solves the underlying issue. I might be misunderstanding, but is the assumption made here: https://github.com/hashicorp/go-plugin/blob/master/log_entry.go#L115 -- that the value part of the KV pairs will always be a string -- valid? AFAIK it's the key that should always be a valid string, not the value.

mitchellh:
Good catch. I'll take a look at this right away. I agree with Jeff that the value should be able to be anything but I'll verify that to be true and pass the right fixups.

Regardless, we shouldn't be doing a cast without an `, ok` check and reporting that failure. Too easy to panic. Sorry!

mitchellh:
Have a fix for the root cause in #40! Thank you!

PR #42, opened 2017-08-16
Fix json detection for log entries

ArthurHlt:
When you have a log entry from the plugin with this kind of format `["elem1", "elem2"]` the actual json detection will detect that it's valid json format (which is actually correct) but the function [parseJSON](https://github.com/hashicorp/go-plugin/blob/master/log_entry.go#L36) expect that the json follow the interface `map[string]interface{}`. 

Obviously, an json array doesn't follow this kind of interface. My little fix is here to detect if the json follow the json object format and not the json array format (replace `json.RawMessage` by `map[string]interface{}` in `isJSON` function)

### DIFF
--- a/client.go
+++ b/client.go
@@ -778,6 +778,6 @@ func (c *Client) logStderr(r io.Reader) {
 }
 
 func isJSON(str string) bool {
-	var js json.RawMessage
+	var js map[string]interface{}
 	return json.Unmarshal([]byte(str), &js) == nil
 }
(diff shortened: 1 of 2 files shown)


mitchellh:
Don't know if this is a coincidence but #43 was also written and that looks to be an approach that removes the need for isJSON entirely. let's go with that! Thank you!

ArthurHlt:
It looks like a coincidence, and yes his solution is actually better. Thank you

PR #51, opened 2018-01-10
Allow the server's logger to be passed in via the ServeConfig

briankassouf:
This will allow a plugin to use go-plugin's logger to pass logs back to the client.

### DIFF
--- a/server.go
+++ b/server.go
@@ -67,6 +67,8 @@ type ServeConfig struct {
 	// relies on this to implement Ping().
 	GRPCServer func([]grpc.ServerOption) *grpc.Server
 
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -66,6 +66,8 @@ type ServeConfig struct {
 	// the gRPC health checking service. This is not optional since go-plugin
 	// relies on this to implement Ping().
 	GRPCServer func([]grpc.ServerOption) *grpc.Server
+
+	Logger hclog.Logger
```
calvn: Include a comment for `Logger` (perhaps mention in there that this is optional).

PR #52, opened 2018-01-11
Add a Connection Broker for GRPC plugins

briankassouf:
This adds a broker object that can serve and dial connections via connection ids. It uses a streaming gRPC service to communicate the connection details between a client and a plugin. Each `AcceptAndServe` call opens a new listener socket and sends the connection info down the stream to the dialer. Since a new connection is opened every call, these calls should be used sparingly. Multiple gRPC server implementations can be registered to a single `AcceptAndServe` call.

Plugins written in another language will, optionally, have to implement the gRPC server/client for the broker, but we provide the proto file for this definition.

Unfortunately there is a slight API breaking change with the `GRPCPlugin` interface. The `GRPCServer` and `GRPCClient` now accept a `GRPCBroker` as the first parameter. Open to ideas on how to make this BC.


briankassouf:
@mitchellh The  `newGRPCBroker` function is here: https://github.com/hashicorp/go-plugin/pull/52/files#diff-d87e36aaf5ae95071a965d9f6d6d40bdR273

In that file is the implementation of the GRPCBroker as well as the server/client used for the gRPC stream.

examples/bidirectional/shared/interface.go
```diff
@@ -0,0 +1,58 @@
...
+	Sum(int64, int64) (int64, error)
+}
+
+// KV is the interface that we're exposing as a plugin.
+type Counter interface {
+	Put(key string, value int64, a AddHelper) error
+	Get(key string) (int64, error)
+}
+
+// This is the implementation of plugin.Plugin so we can serve/consume this.
```
calvn: Should be plugin.GRPCPlugin instead of plugin.Plugin

grpc_broker.go
```diff
@@ -0,0 +1,455 @@
...
+
+	// Proccess send stream
+	go func() {
+		for {
+			select {
+			case <-doneCh:
+				return
+			case <-s.quit:
+				return
+			case s := <-s.send:
```
calvn: Minor nit, but perhaps rename `s` into something else, since `s` initially refers to `*gRPCBrokerServer`, but after the assignment it's a `*sendErr`. Even though there's not conflict in overriding `s` in this case, it makes it somewhat harder to read. Same for `StartStream` on the client side.

PR #53, opened 2018-01-16
Don't close the listener in TestPluginGRPCConn

jbardin:
The grpc transport is responsible for managing connections, and may
create new connections as required. When cycling servers and connections
in quick succession in tests, a new connection is occasionally created
to service a request.

The grpc server must be stopped manually by callers of the `TestPluginGRPCConn` function.

Expose the `Stop` and `GracefulStop` methods from the `grpc.Server` so they can be called outside of the plugin package. 

Since this was found due to intermittently failing tests in Terraform using streams, I also added a test that uses a bidirectional grpc streaming service.


jefferai:
@jbardin note that there are a number of conflicts

jbardin:
Thanks, I'll rebase

jbardin:
rebased again ;)

grpc_server.go
```diff
@@ -90,6 +90,16 @@ func (s *GRPCServer) Init() error {
 	return nil
 }
 
+// Stop calls Stop on the underlying grpc.Server
+func (s *GRPCServer) Stop() {
```
mitchellh: Any reason to not just call this `Close() error` to implement `io.Closer`? (This or graceful)

jbardin: I thought about `Close`, but ended up using `Stop` because that's what the grpc server calls it, and to keep the similarity between `Stop` and `StopGraceful`. 

I don't think fulfilling `io.Closer` is a string reason, since it doesn't fulfill any other `io` interfaces, but I'm not opposed to naming it differently.

PR #55, opened 2018-01-18
Provide a context for managing the lifecycle of gRPC plugins

briankassouf:
If a plugin server goes away, gRPC clients will wait in the connecting and transient_failure states for some time (see https://github.com/grpc/grpc/blob/master/doc/connectivity-semantics-and-api.md). This PR provides a context to the host process that will be canceled when go-plugin detects the plugin process has exited. If the host process detects the context has been closed it can take steps to recover.


client.go
```diff
@@ -457,6 +462,8 @@ func (c *Client) Start() (addr net.Addr, err error) {
 
 			// Close the logging channel since that doesn't work on reattach
 			close(c.doneLogging)
+			// Cancel the context
```
mitchellh: Nitpick to add a blank line above here

PR #58, opened 2018-01-23
Fix small typo in gRPC plugin error message

evan2645:
Small cosmetic fix.

For what its worth, I couldn't build or run the tests because there's no vendor dir and there doesn't seem to be any dep management, so I was missing some stuff.


mitchellh:
Thanks!

mitchellh:
On running tests: this is a library, and it is customary for libraries in Go to not include vendor directories (only edge applications consuming libraries have a vendor dir). You're expected to `go get` the dependencies.

PR #62, opened 2018-02-03
Remove "log" package

### LINKED ISSUE #44: CleanupClients() does not use hclog
The CleanupClients method is not defined on a type, and thus does not have direct access to a configured logger. It logs a single line using stdlib `log`:
https://github.com/hashicorp/go-plugin/blob/a5174f84d7f8ff00fb07ab4ef1f380d32eee0e63/client.go#L235

This is problematic because 1) it is not structured or formatted as we would expect it to be, and 2) it goes to stdout regardless of the actual logger configuration.

The package does manage a list of instantiated clients, each of which has a logger defined on its struct. I can think of two solutions:
* A series of "waiting" and "stopped" messages, specific to each plugin, logged through its own logger
* Pick the first client, if any, and log the single message through there.

Thoughts on the best option? Perhaps you can think of other options? Happy to send a patch assuming we are aligned

mitchellh:
Fixes #44 

Remove the last usage of the standard `log` package so we're all `hclog` now.

This removes a log statement that isn't recreated anywhere else. I tried putting the statements into `client.Kill` but it gets pretty noisy. I also realized that clients aren't named right now so if you launch a bunch of plugins you invariably get a lot of untargeted "Kill requested" type logs which aren't helpful.

I think it may be better to just remove this for now.

### DIFF
--- a/client.go
+++ b/client.go
@@ -10,7 +10,6 @@ import (
 	"hash"
 	"io"
 	"io/ioutil"
-	"log"
 	"net"
 	"os"
 	"os/exec"
@@ -234,7 +233,6 @@ func CleanupClients() {
 	}
 	managedClientsLock.Unlock()
 
-	log.Println("[DEBUG] plugin: waiting for all plugin processes to complete...")
 	wg.Wait()
 }
 


JulienBreux:
@jbardin / @briancain When this PR can be Merge?

jefferai:
How about now? :-)

PR #63, opened 2018-02-06
Make the tests pass the race detector

jefferai:
These changes are relatively obvious; the one that might cause a little
heartburn is declaring a package-level variable for test stdout/err, but
given the control flow of the test functions it's either that or modify
the test function APIs. I'm happy to do the latter if it's deemed to not
be a big deal, since they're test functions.

### DIFF
--- a/testing.go
+++ b/testing.go
@@ -3,11 +3,17 @@ package plugin
 import (
 	"bytes"
 	"context"
+	"io"
 	"net"
 	"net/rpc"
 
 	"github.com/mitchellh/go-testing-interface"
 	"google.golang.org/grpc"
+)
+
+var (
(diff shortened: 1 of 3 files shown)


testing.go
```diff
@@ -3,13 +3,19 @@ package plugin
...
 	"net"
 	"net/rpc"
 
 	"github.com/mitchellh/go-testing-interface"
 	"google.golang.org/grpc"
 )
 
+type TestOptions struct {
+	ServerStdout io.ReadCloser
+	ServerStderr io.ReadCloser
```
mitchellh: It would be preferrable to have doc comments on these.

jefferai: Fair enough, will do that.

PR #65, opened 2018-02-16
fix broken GRPC bidirectional example

ghouscht:
This PR fixes the GRPC bidirectional example in examples/. The import paths were wrong and  the plugin.GRPCPlugin interface was not implemented properly thus the example was not usable as documented in the README.


mitchellh:
Thanks!

PR #68, opened 2018-04-03
client: don't name the logger

briankassouf:
The caller should be able to specify the logger name without go-plugin adding to it. This is useful in vault since we have a name that each plugin is registered under, and we'd rather use that name than the filepath. 

If we are worried about this change I can instead make it a setting in the client config struct.

### DIFF
--- a/client.go
+++ b/client.go
@@ -13,7 +13,6 @@ import (
 	"net"
 	"os"
 	"os/exec"
-	"path/filepath"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


jefferai:
LGTM, pending others' views about making it a client config option.

calvn:
I like the idea of having it as a config option, with the binary's name as the fallback/default if none is provided. If you are running multiple plugins, I think it's helpful to see which log is coming from which plugin, so I'd rather see the name of the binary rather than nothing if this parameter was not configured by the caller.

jefferai:
That makes sense to me.

mitchellh:
Hey Brian, Jeff, if you want to open another PR with the given feedback, let's do that. I don't think we're merging this?

PR #77, opened 2018-08-26
Add go.mod

mitchellh:
This adds a go.mod file, automatically made using:

    $ go mod init
    $ go test

This doesn't add any versioning yet to this module, we can tag it v1 after if you're comfortable, but I want to double check deps first. 

There are some HashiCorp deps that aren't versioned yet in the go.mod file (hclog and so on). I have separate PRs for those and will tag those later.


dchenk:
Thank you for this! :)

mitchellh:
I think this is good, fairly harmless for now without the tag.

PR #81, opened 2018-10-02
make sure Kill is idempotent

jbardin:
The Kill method is documented as safe to call multiple times, but because the process handle is not removed after the first time, it results in logged errors. 

In troubleshooting this, it also turns out that the child process is never gracefully terminated. I had to remove that from the scope of this PR due to time constraints (and the lack of a simple cross-platform solution), but there is now a log message and associated comment in the source. A POSIX solution is quite simple, with a signal handler to call Stop on the GRPCServer, but that doesn't leave a good solution for Windows. Another option is to implement a control protocol over the plugin's Stdin stream, but that can be left for further discussion.

### DIFF
--- a/client.go
+++ b/client.go
@@ -358,10 +358,18 @@ func (c *Client) Kill() {
 	doneCh := c.doneLogging
 	c.l.Unlock()
 
-	// If there is no process, we never started anything. Nothing to kill.
+	// If there is no process, there is nothing to kill.
 	if process == nil {
 		return
 	}
+
+	defer func() {
+		// Make sure there is no reference to the old process after it has been
+		// killed.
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -392,8 +400,11 @@ func (c *Client) Kill() {
 	if graceful {
 		select {
 		case <-doneCh:
+			// FIXME: we can never reach here, because the plugin process is
+			// never signaled to exit.
```
mitchellh: I believe this can happen in cases where the plugin exited on its own (probably not gracefully but on its own).

jbardin: good point, I'll reword that

PR #86, opened 2018-10-30
Fix package name, set to go_plugin

th0br0:
The package path is currently wrong and causes errors when compiling via e.g. bazel:
`process.go:1:9: package plugin; expected go_plugin`


mitchellh:
Thanks for the contribution! 

It appears however this is an issue in Bazel's expectations. Go package names _do not_ have to match the folder name. While the common case is that they do match the name, in a handful of cases they do not. 

Changing the package name would be a much larger breaking change and would break all downstream consumers who do not explicitly rename the package with an alias.

And finally, underscores in package names are valid but not idiomatic in Go source. They certainly exist and we use them from time to time but its not common and we'd prefer not to bring that pattern into our source.

th0br0:
Thanks for the detailed explanation! Not being a 'native Gopher' I really appreciate it.

I guess this boils down to wrong expectations on my end.

PR #88, opened 2018-11-30
grpc shutdown and child process handling

jbardin:
This PR's main goal is to fix the shutdown process of grpc plugins, which would always timeout.
In the process we also clean up the subprocess output handling, which speeds up the start/stop of the child process. This broke some tests that were dependent on the shutdown taking more time, but those were patched with a slight delay and can be fixed with a more robust solution at a later time.

The grpc server couldn't exit at all, because the Serve method was waiting on a channel that was never closed. The Serve method is what should be closing the channel to notify when the server is done. 

The close signal is implemented by adding a small grpc service (`GRPCController`) with a `Shutdown` method to indicate that the process should exit. There is still something blocking a `GracefulStop` attempt, but the implementation can be extended to try a graceful stop with a timeout once that is resolved.

The GRPCController and GRPCBroker protobuf definitions were moved into an internal package, which prevents the generated code from polluting the public API.


jbardin:
Yes, let me get a failing test for the before case. I don't like the sleeps either, but they were just inserted until I had time to refactor the the tests (which I do now that the next TF alpha is out).

I'll update the tests in this PR before merging.

jbardin:
Rebased on the latest master.  This is closer now, but I've still managed to induce some odd ordering during shutdown, so there will be at least one more commit incoming.

jbardin:
All set now. Cleaned up the temporary commits in the middle, and all tests all pass under `-race` too 🏎 🏎

jbardin:
Sorry, the failing test change was squashed into the rest of the test fixes, since it was pretty ugly to rebase in before the rest of the commits. Any of the grpc tests that call `killed()` will fail prior to the shutdown changes.

PR #105, opened 2019-03-19
Add WaitGroup to bidi example to wait for grpc.Server init

lukeb2e:
The bidirectional example currently does not work.

This change makes the example work again by adding a WaitGroup.

Kind regards,
Luke


ghostsquad:
It would be great if the examples worked. Could this get merged?

PR #106, opened 2019-03-22
Log unexpected json, and set levels for plain text logs

jbardin:
Don't drop json output if it's missing a log level. Log the original
line rather than as key-value data, since it likely wasn't intended for
hclog. Even if plugins try to avoid sending raw json objects, it's hard
to avoid if they are writing strings containing newlines (e.g. http
responses), which will be broken into multiple messages by the plugin
client.

Most plugins using go-plugin were previously using the convention of
prefixing log lines with the log level as `[LEVEL]`. Use these prefixes
if available to set the log level.

### DIFF
--- a/client.go
+++ b/client.go
@@ -989,6 +989,11 @@ func (c *Client) logStderr(r io.Reader) {
 				l.Warn(entry.Message, out...)
 			case hclog.Error:
 				l.Error(entry.Message, out...)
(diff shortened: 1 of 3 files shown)


jbardin:
Thanks @elonlee! those arguments do seems to be reversed 😞

PR #110, opened 2019-04-15
More informative errors for serverListener_tcp

arykalin:
Fixing issue 107 - https://github.com/hashicorp/go-plugin/issues/107

Issue #107, opened 2019-03-30
Report serverListener_tcp() errors more clearly


j0hnwhyte:
I'm trying to run the system on Windows, and the way it uses to connect to plugins is TCP. I see that the functions refer to two environment variables, `PLUGIN_MIN_PORT` and `PLUGIN_MAX_PORT`. If they are undefined, a clueless error is returned.

I kind of understand why they have to be declared as environment variables, it makes good sense, though the error took me some time to find out what it actually was. Please report the absence of these variables more clearly. Or make it configurable in the code.

Thanks

arykalin:
We need to check if variables are set, otherwise we have very strange error:
2019-04-09T18:56:31.250+0300 [DEBUG] rollback: attempting rollback: path=secret/
2019-04-09T18:56:58.435+0300 [WARN]  received plugin exited before we could connect attempting as db plugin, attempting as auth/secret plugin
2019-04-09 18:56:58.581346 I | [ERR] plugin: plugin acceptAndServe error: strconv.ParseInt: parsing "": invalid syntax
2019-04-09T18:57:31.250+0300 [DEBUG] rollback: attempting rollback: path=auth/token/

mitchellh:
This is fixed by #111!

PR #111, opened 2019-04-15
listener/tcp: better handling of min/max port env vars

calvn:
This PR addresses the case where `PLUGIN_MIN_PORT` and/or `PLUGIN_MAX_PORT` are not set and thus treated as empty strings. In this case, `strconv.ParseInt` will error and and the plugin will fail to run.

This PR sets those values as 0 if not provided and lets the OS handle the port assignment (by effectively giving it an address of `127.0.0.1:0`). It also handles the case where the min port might be provided as a value larger than the max port, which would skip listener instantiation and error with "Couldn't bind plugin TCP listener".

Related to https://github.com/hashicorp/vault/issues/6587

### DIFF
--- a/docs/internals.md
+++ b/docs/internals.md
@@ -61,3 +61,16 @@ Where:
     is omitted (older versions), this is "netrpc" for Go net/rpc. This can
     also be "grpc". This is the protocol that the plugin wants to speak to
     the host process with.
+
+## Environment Variables
+
+When serving a plugin over TCP, the following environment variables can be
+specified to restrict the port that will be assigned to be from within a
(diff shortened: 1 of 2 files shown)


kalafut:
Also related: https://github.com/hashicorp/go-plugin/issues/107

calvn:
Also added the error wrapping from #110 to this PR. Apologies in advance for hijacking that PR!

arykalin:
Hi. Is there any information then it will be merged?

tr1ck3r:
@briankassouf is there a target release for this fix?  This issue is currently preventing people from being about to use the PKI plugins that Venafi has developed for Vault.

PR #113, opened 2019-04-22
Add GRPC Stdio Copying system

sargun:


jefferai:
A note to any reviewers, some of the comments indicate that this isn't necessarily fully implemented.

grpc_client.go
```diff
@@ -67,6 +74,32 @@ func newGRPCClient(doneCtx context.Context, c *Client) (*GRPCClient, error) {
...
+		response, e := stream.Recv()
+		switch e {
+		case nil:
+		case io.EOF, context.Canceled:
+			return
+		default:
+			logger.Error("Experienced issue in gRPC copier", "error", e)
+			return
+		}
+		writer.Write(response.Data)
```
yurishkuro: this would be better placed inside `case nil:`

yurishkuro:
I wonder, wouldn't it be easier for the client to simply read the stdout/err of the child process, split on `\n` and forward to its own stdout/err? This way the plugins won't need to implement anything special, especially when they are written in other languages.

mitchellh:
Superseded by #135. Thank you!

PR #117, opened 2019-05-08
increase the default grpc send/recv message size

ming-relax:


jefferai:
Is there a reason not to use the math.MaxInt32 instead?

PR #118, opened 2019-05-20
ensure we finish reading stderr pipe before calling cmd.Wait()

### LINKED ISSUE #116: reading plugin stderr read |0: file already closed
I'm seeing intermittent errors being logged from go-plugin with a message of "reading plugin stderr", and an error of "read |0: file already closed".  This seems to happen when plugins are terminated.

In looking at https://golang.org/src/os/exec/exec.go?s=17633:17682#L601, and `client.go`, it seems like there may be a race condition where the reader side of the pipe is closed after the command exits, but the go-plugin client is still trying to read from it.  The client's `logStderr` is looking for an `io.EOF` error, but in this case I think it's actually getting some variant of the `os.ErrClosed` error. I'm not familiar enough with the code to understand exactly what's going on/what should be changed, though. Hope this can point someone in the right direction!

skriss:
Signed-off-by: Steve Kriss <krisss@vmware.com>

Fixes #116 

cc @jbardin

### DIFF
--- a/client.go
+++ b/client.go
@@ -87,6 +87,10 @@ type Client struct {
 	// goroutines.
 	clientWaitGroup sync.WaitGroup
 
+	// stderrWaitGroup is used to prevent the command's Wait() function from
+	// being called before we've finished reading from the stderr pipe.
+	stderrWaitGroup sync.WaitGroup
+
 	// processKilled is used for testing only, to flag when the process was
 	// forcefully killed.
 	processKilled bool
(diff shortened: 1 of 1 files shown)


skriss:
FYI, I let this run overnight in a daemon process and can confirm I'm no longer getting the error from #116.

skriss:
@jbardin just pinging on this, thanks :)

skriss:
Another ping on this @jbardin.

jbardin:
Sorry about the delay @skriss, we've been tied up with the terraform side of things. 
I'll take care of this shortly, so we can get it in an upcoming release over there too.

skriss:
No worries, I figured :) Thanks!

PR #127, opened 2019-10-25
Start the plugin using the configured stdin, if available

carolynvs:
## What is the change?
When the `Cmd` struct already has `stdin` already configured, use that instead of `os.Stdin` when starting a plugin.

## Why?
I wanted to pass data to my plugins to assist with initialization, passing one-time configuration data that isn't part of the interface used to communicate with the plugin. I was having trouble getting this sent properly via `os.Stdin` honestly... 😊 

This seems like an improvement because it doesn't require remembering the original `os.Stdin` and swapping it back, and makes it easier for the host to pass a different stdin to a plugin, or mock it out for a test.


ghostsquad:
Any progress on this?

PR #128, opened 2019-10-31
Update to latest protobuf and gRPC

syndbg:
Just updates by default to the latest protobuf and gRPC. 

Follow up of #123 , cause in that PR it's already behind few versions.

Rationale:

* More up to date defaults, nothing else


jefferai:
Please stick to one PR. You can always update it with more commits.

syndbg:
@jefferai The other is not my PR

PR #129, opened 2019-11-05
Simplify string to int.

d-kuro:
It is simpler to use `strconv.Atoi(s)` than `strconv.ParseInt(s, 10, 0)`
https://golang.org/pkg/strconv/#Atoi


mitchellh:
Thanks this seems safe and the tests pass which do exercise this.

PR #133, opened 2020-01-07
Create ServeAsync

appilon:
Callers should block with the returned wait() func. This is useful for type asserting to `plugin.GRPCServer`, which gives the caller access to `GracefulStop` and `Stop`. This would be helpful for the Terraform plugin SDK so we can propagate cancellation naturally though gRPC's `context.Contexts`
```
server, wait := plugin.ServeAsync(opts)
someService.OnStop = func() {
  server.(*plugin.GRPCServer).GracefulStop()
}
wait()


appilon:
closing in favor of #134

PR #134, opened 2020-01-08
Create ServeWithContext

appilon:
This feature allows a GRPC server to be manually cancelled via the context. This context does not override/get set as the root context of the GRPC Server so it can't be used for setting server-default metadata, it is currently only used for cancellation signalling. After the context cancels `ServeWithContext` still blocks until the server has completely exited.

A test case was written following the precedent set by the `cleanup` test, writing out a file proving to the TestCase the plugin closed as a result of manual cancellation.


appilon:
Supporting this feature would be an antipattern after some discussion.

PR #135, opened 2020-02-24
gRPC stdout/stderr syncing support

### LINKED ISSUE #93: stdout/stderr sync not implemented for gRPC plugins
I might be missing something completely, but it seems that the stdout/stderr sync feature is not implemented for the GRPC server path. 

The [`GRPCServer.Stdout` and `GRPCServer.Stderr` fields](https://github.com/hashicorp/go-plugin/blob/master/grpc_server.go#L51-L52) are not used anywhere. They are set up however in the same way as in `RPCServer` in the common codepath in [`Serve()`](https://github.com/hashicorp/go-plugin/blob/master/server.go#L308-L309) it's just RPCServer has control over the channel, while in GRPC I guess it is not that obvious how the sync can be implemented. Is my understanding correct? 

I think this asymmetry should be documented between netRPC and gRPC! Are there any more differences? 

Logs in gRPC plugins can work using the `log`, `hclog` packages, they _are_ pointing to the old stderr.   

Workaround for getting the output of `fmt.Println()` for example to `stdErr` involves saving the old `stdErr` in a var and using that (for some reason the same trick does not work with `stdOut`):

````
var oldOut *os.File
var oldErr *os.File
// Here is a real implementation of KV that writes to a local file with
// the key name and the contents are the value of the key.
type KV struct{}

func (KV) Put(key string, value []byte) error {
	fmt.Fprintf(oldOut,"Hello, I'm the plugin putter.1\n") # this does not work
	fmt.Fprintf(oldErr,"Hello, I'm the plugin putter.2\n") # this works
	fmt.Fprintf(os.Stderr,"Hello, I'm the plugin putter.3\n") # this does not work
	fmt.Fprintf(os.Stdout,"Hello, I'm the plugin putter.4\n") # this does not work
	log.New(os.Stderr, "hihii", 0).Printf("test") # this does not work
	log.Printf("log hello:) - %v %v vs %v %v\n", os.Stdout, os.Stderr, oldOut, oldErr) # this works
	value = []byte(fmt.Sprintf("%s\n\nWritten from plugin-go-grpc", string(value)))
	return ioutil.WriteFile("kv_"+key, value, 0644)
}

func (KV) Get(key string) ([]byte, error) {
	return ioutil.ReadFile("kv_" + key)
}

func main() {
	oldOut = os.Stdout
	oldErr = os.Stderr
	plugin.Serve(&plugin.ServeConfig{
		HandshakeConfig: shared.Handshake,
		Plugins: map[string]plugin.Plugin{
			"kv_grpc": &shared.KVGRPCPlugin{Impl: &KV{}},
		},
		// A non-nil value here enables gRPC serving for this plugin...
		GRPCServer: plugin.DefaultGRPCServer,
	})
}
```` 

In comparison the Sync feature does work well in the netRPC plugin when `SyncStdout` and `SyncStderr` are set on the `plugin.ClientConfig`.

### PR DESCRIPTION
**Pull request by mitchellh**
Fixes #93 

This is a feature that has existed in go-plugin for a few years by
setting SyncStdout, SyncStderr on ClientConfig, but this feature only
worked with net/rpc plugins previously.

This feature will sync the os.Stdout/Stderr output from the plugin to
the specified writers. If no writers are specified, it drops the data.
Note that the os.Stdout/Stderr that is synced is the overridden writers
that you can see set in server.go around line 250 at the time of this
commit.

The use case for this feature is that a plugin that subprocesses for
example can connect stdout/stderr and the output of the subprocess will
reach the target synced stream (which may just be the host process
stdout/stderr).

## Backwards Compatibility

Should be preserved. In `grpc_client.go` we noop the stdio syncing if we get a `codes.Unavailable` when trying to connect to the stdio service. This will make existing plugins that don't serve this service work fine. 

## Implementation Notes

This is implemented via a fairly naive "stdio" gRPC service. We chunk the data at 1KB boundaries currently because in practice stdout/stderr is pretty quiet. We don't expect so much data ever that this will be a problem. If it is, we can change this.

There are various improvements we _could_ make but I don't think matter: right now even if syncing is disabled, all plugins will still transport all stdout/stderr to the host process. Most plugins don't have ANY stdout/stderr so this shouldn't matter. I didn't do this because net/rpc doesn't do this and I mimicked that implementation.

### CODE REVIEW

**Comment by mitchellh** [comment]
And thanks to #113. This is very similar and inspired by the impl., I changed it a bit and added tests.

**Comment by yurishkuro** [comment]
@mitchellh could you please help my understand (also this https://github.com/hashicorp/go-plugin/pull/113#issuecomment-586667462) why GRPC needs to be involved in shipping stdout/err back to parent process, instead of using Cmd.StdoutPipe/StderrPipe?

**Comment by mitchellh** [comment]
> @mitchellh could you please help my understand (also this [#113 (comment)](https://github.com/hashicorp/go-plugin/pull/113#issuecomment-586667462)) why GRPC needs to be involved in shipping stdout/err back to parent process, instead of using Cmd.StdoutPipe/StderrPipe?

Yeah sure! We use the real process stdout/stderr (i'll call them FD1 and FD2 so we don't get mixed up with the "synced" stdout/stderr) for internal purposes:

* FD1 (stdout) is used as the [plugin handshake stream](https://github.com/hashicorp/go-plugin/blob/master/docs/internals.md#handshake). In theory we could use it as a normal stdout AFTER the handshake and sync it back but we want to retain the ability to do more command & control there in the future. 

* FD2 (stderr) is used as the log output. Anything on FD2 is re-logged in the host process logger instance. This simplifies debugging. There isn't an easy way to repurpose this for "normal" stderr without multiplexing output (i.e. prefixing it or something) which is kind of nasty.

So what we do instead is [overwrite os.Stdout/Stderr](https://github.com/hashicorp/go-plugin/blob/809113480b559c989ea9cfcff62e9d387961f60b/server.go#L349-L350) in `plugin.Serve` to go to the pipes that get streamed back -- in this case over a gRPC service and in net/rpc's case a Yamux stream. 

Is it possible to repurpose FD1/FD2 for stdout/stderr syncing? Probably, but I think it would introduce a lot more complexity and risk destabilizing our plugin system much more than this approach.

**Comment by yurishkuro** [comment]
Thanks, makes sense now. Agree that it's not worth complicating the handshake protocol at this point (although if plugins log a lot, it probably would've been more efficient to reuse stdio streams).

> FD2 (stderr) is used as the log output. Anything on FD2 is re-logged in the host process logger instance. 

I guess the issues we've seen some people having with non-Go implementation of plugins for Jaeger were due to them writing logs to stdout, not stderr.

File: internal/plugin/grpc_stdio.proto [hunk 1]
```diff
@@ -0,0 +1,29 @@
...
+  //
+  // Callers should connect early to prevent blocking on the plugin process.
+  rpc StreamStdio(google.protobuf.Empty) returns (stream StdioData);
+}
+
+// StdioData is a single chunk of stdout or stderr data that is streamed
+// from GRPCStdio.
+message StdioData {
+  enum Channel {
+    STDOUT = 0;
```
paultyng: Probably not a big deal, but should zero here be `UNSPECIFIED` to make it non-optional to set and follow the normal style (this convention may be dated to v2, I'm not sure)?

paultyng: I guess optionally you could just skip zero.

grpc_stdio.go
```diff
@@ -0,0 +1,196 @@
...
+		case plugin.StdioData_STDERR:
+			w = stderr
+
+		default:
+			c.log.Warn("unknown channel, dropping", "channel", data.Channel)
+			continue
+		}
+
+		// Write! In the event of an error we just continue.
+		c.log.Trace("received data", "channel", data.Channel.String(), "len", len(data.Data))
```
jbardin: These `waiting` and `received` log messages within the receive loop feel like they might be too chatty for busy clients.

grpc_stdio.go
```diff
@@ -0,0 +1,196 @@
...
+	for {
+		// Make our data buffer. We allocate a new one per loop iteration
+		// so that we can send it over the channel.
+		var data [1024]byte
+
+		// Read the data, this will block until data is available
+		n, err := bufsrc.Read(data[:])
+
+		// If we hit EOF we're done copying
+		if err == io.EOF {
```
jbardin: Data should be sent before this check

grpc_stdio.go
```diff
@@ -0,0 +1,196 @@
...
+		if n == 0 {
+			continue
+		}
+
+		// We have data! Send it on the channel. This will block if there
+		// is no reader on the other side. We expect that go-plugin will
+		// connect immediately to the stdio server to drain this so we want
+		// this block to happen for backpressure.
+		dst <- data[:n]
+	}
```
jbardin: This function isn't checking for any errors other than `EOF`, which would drop into a busy loop

internal/plugin/grpc_stdio.proto
jbardin: I also like that, only to align stdout and stderr with their default FDs, just for some sense of superfluous consistency ;)

grpc_stdio.go
mitchellh: The stdlib docs state that there will never be any data on `io.EOF` so we don't need to send anything (`n == 0` guaranteed). Am I misunderstanding?

grpc_stdio.go
mitchellh: Good call. Changed this to log at a warning level and exit.

grpc_stdio.go
mitchellh: I changed this to a `c.log.IsTrace()` check prior. We don't normally ever default trace logs (even in dev I think we normally do Debug level), and I've been using them more since in a pinch of debugging they've been helpful.

I'm fine removing this but felt it could be useful in debugging why stdio might not be working if that ever happened cause you can see on both sides the data receival/send.

internal/plugin/grpc_stdio.proto
mitchellh: I'll set a zero value here, I usually do that :)

mitchellh: Done (not you do have to explicitly set a zero value or `protoc` errors)

grpc_stdio.go
mitchellh: Note I did update to send data before checking `err != nil` AFTER this though, since the above guarantee is only made on EOF. I guess it doesn't hurt to do it anytime but let me know what you think.

mitchellh:
Addressed feedback.

grpc_stdio.go
```diff
@@ -0,0 +1,208 @@
...
+	ctx context.Context,
+	log hclog.Logger,
+	conn *grpc.ClientConn,
+) (*grpcStdioClient, error) {
+	client := plugin.NewGRPCStdioClient(conn)
+
+	// Connect immediately to the endpoint
+	stdioClient, err := client.StreamStdio(ctx, &empty.Empty{})
+
+	// If we get an Unavaiable error, this means that the plugin isn't
```
kalafut:
```suggestion
	// If we get an Unavaliable error, this means that the plugin isn't
```

grpc_stdio.go
jbardin: Since this is a generic `io.Reader`, we should follow the broadest form of the spec. I actually can't recall which concrete types behave which way, since we always need to cover them all in read loops like this. The portion of the stdlib that documents this is in `io.Reader` (with emphasis on why I always check this in review): 

> When Read encounters an error or end-of-file condition after successfully reading n > 0 bytes, it returns the number of bytes read. _It may return the (non-nil) error from the same call_ or return the error (and n == 0) from a subsequent call. An instance of this general case is that a Reader returning a non-zero number of bytes at the end of the input stream may return either err == EOF or err == nil. The next Read should return 0, EOF.
> 
> _Callers should always process the n > 0 bytes returned before considering the error err_. Doing so correctly handles I/O errors that happen after reading some bytes and also both of the allowed EOF behaviors.

mitchellh: Sounds good. It isn't a generic reader though since I always wrap it in `bufio.Reader`. But I'll just check the bytes earlier, no harm no foul!

mitchellh:
Whoops sorry @paultyng I re-requested your review but meant @jbardin. :)

marefr:
@mitchellh regarding backwards compatibility I don't agree that works as expected/explained above. Already merged #139 plus a new one opened #142, but those are more patching the issues. Maybe we need to rethink what error codes to check for identifying old plugin?

Have you been able to reproduce the behavior of getting back an `Unavailable` error [here](https://github.com/hashicorp/go-plugin/blob/991a34ba3526019b82938e1c292e1fe4b5782f43/grpc_stdio.go#L106) or [here](https://github.com/hashicorp/go-plugin/blob/991a34ba3526019b82938e1c292e1fe4b5782f43/grpc_stdio.go#L135) 

Let me know if rather would like me to open an issue about it.

mitchellh:
I did test `Unavailable` prior to merge. If you remove that check, old tests will fail. They probably won't anymore since we have some additional checks now but that was the first one that went in.

PR #138, opened 2020-03-13
Register grpc reflection service

### LINKED ISSUE #137: Add optional support for gRPC Reflection
I don't believe the gRPC server is exposed to calling applications, so we are unable to [turn on reflection](https://github.com/grpc/grpc-go/blob/master/Documentation/server-reflection-tutorial.md#enable-server-reflection).

The diff example from the gRPC docs, walks through the required changes:

```diff
--- a/examples/helloworld/greeter_server/main.go
+++ b/examples/helloworld/greeter_server/main.go
@@ -40,6 +40,7 @@ import (
        "google.golang.org/grpc"
        pb "google.golang.org/grpc/examples/helloworld/helloworld"
+       "google.golang.org/grpc/reflection"
 )

 const (
@@ -61,6 +62,8 @@ func main() {
        }
        s := grpc.NewServer()
        pb.RegisterGreeterServer(s, &server{})
+       // Register reflection service on gRPC server.
+       reflection.Register(s)
        if err := s.Serve(lis); err != nil {
                log.Fatalf("failed to serve: %v", err)
        }
```

paultyng:
Closes #137 

Similar to our usage of the standard gRPC health service, also register the standard reflection service.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/golang/protobuf v1.3.4
 	github.com/hashicorp/go-hclog v0.0.0-20180709165350-ff2cf002a8dd
 	github.com/hashicorp/yamux v0.0.0-20180604194846-3520598351bb
+	github.com/jhump/protoreflect v1.6.0
 	github.com/mitchellh/go-testing-interface v0.0.0-20171004221916-a61a99592b77
 	github.com/oklog/run v1.0.0
 	github.com/stretchr/testify v1.3.0 // indirect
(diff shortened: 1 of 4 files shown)


mitchellh:
This seems good but I think I'd add a test as well that uses the reflection client to verify that it is listening properly.

paultyng:
Added a test, the package `github.com/jhump/protoreflect` just has some convenience wrapping to interact with the reflection service (its used in [grpcurl](https://github.com/fullstorydev/grpcurl)) and only referenced in the tests here, so not a runtime dependency.

PR #139, opened 2020-04-07
Handle stdio service not implemented

marefr:
Fixes regression introduced by #135 when client uses latest version of this package and plugin is compiled with earlier version.

Before this fix, my client log were flooded with the following messages:
```
DBUG[04-07|14:04:20] waiting for stdio data
EROR[04-07|14:04:20] error receiving data err="rpc error: code = Unimplemented desc = unknown service plugin.GRPCStdio"
```

With these fixes I only receive the following warning message when plugin starts:
```
DBUG[04-07|14:04:20] waiting for stdio data
WARN[04-07|14:04:20] received EOF, stopping recv loop  err="rpc error: code = Unimplemented desc = unknown service plugin.GRPCStdio"
```

### DIFF
--- a/grpc_stdio.go
+++ b/grpc_stdio.go
@@ -7,12 +7,11 @@ import (
 	"io"
 
 	empty "github.com/golang/protobuf/ptypes/empty"
-	"github.com/hashicorp/go-hclog"
+	hclog "github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-plugin/internal/plugin"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
-
-	"github.com/hashicorp/go-plugin/internal/plugin"
 )
 
 // grpcStdioBuffer is the buffer size we try to fill when sending a chunk of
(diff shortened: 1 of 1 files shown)


grpc_stdio.go
```diff
@@ -101,10 +100,10 @@ func newGRPCStdioClient(
 	// Connect immediately to the endpoint
 	stdioClient, err := client.StreamStdio(ctx, &empty.Empty{})
 
-	// If we get an Unavailable error, this means that the plugin isn't
+	// If we get an Unavailable or Unimplemented error, this means that the plugin isn't
 	// updated and linking to the latest version of go-plugin that supports
 	// this. We fall back to the previous behavior of just not syncing anything.
-	if status.Code(err) == codes.Unavailable {
+	if status.Code(err) == codes.Unavailable || status.Code(err) == codes.Unimplemented {
```
marefr, line 106: Still a bit confused of this check since an error was never returned from line 101 above in my case.

grpc_stdio.go
```diff
@@ -135,6 +134,7 @@ func (c *grpcStdioClient) Run(stdout, stderr io.Writer) {
 			if err == io.EOF ||
 				status.Code(err) == codes.Unavailable ||
 				status.Code(err) == codes.Canceled ||
+				status.Code(err) == codes.Unimplemented ||
```
marefr, line 137: This was the actual fix that resolved the flood of messages in my log.

mitchellh:
Interesting. LGTM though thank you.

PR #144, opened 2020-05-08
Add ServeOption to disable syncing stdout.

paddycarver:
By default, os.Stdout and os.Stderr are overwritten with handlers that
pass their output over gRPC to the client. When the server is running as
the main process, or running in the same process as the client, this has
unfortunate side-effects.

In this case, running a server in the TestMain process of `go test`
swallows some test output, include coverage information, which isn't
ideal.

This adds an option to disable this behavior when starting a server.

### DIFF
--- a/server.go
+++ b/server.go
@@ -98,6 +98,8 @@ type ServeConfig struct {
 	// complete. The caller should NOT close this listener once `Serve` is
 	// called.
 	Listener net.Listener
+
+	DisableStdoutSync bool
 }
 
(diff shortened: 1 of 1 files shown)


paultyng:
Alternatively I think we could just `tee` the output to retain outputting the the original `os.Stdout` and `os.Stderr` here: https://github.com/hashicorp/go-plugin/blob/b3be083a51a6d268ea975812a51ae6289ef1c3ba/server.go#L391-L393

paultyng:
Superseded by #145

PR #145, opened 2020-05-13
Add a test mode to enable testing plugins with `go test` more easily, Delve integration, etc.

mitchellh:
This is a special configuration on the plugin server (the plugin side)
that enables running the plugin in-process and eases many of the
behaviors around that.

The desire to do this is for more easily using `go test` (and similar)
with plugins. This mechanism allows the plugin serving to start up
in-process, to grab a ReattachConfig, and to be able to send that
ReattachConfig to some other plugin client for connection.

Behaviorally, this prevents plugin serving from exiting the process with
os.Exit, prevents taking over os.Stdout in such a hostile way, prevents
connection info going to os.Stdout which floods the test logs, etc.

## Usage

The expected usage can sort of be seen in our own test that I wrote
for this feature. See `TestServer_testMode`. You'd start the plugin 
in test mode in the same process as your test. Then you grab the 
reattach config and use that to connect.

In Terraform's case, the plan is to subprocess out to `terraform`
and set a special variable or flag with the reattach info. That's fine, too.

## Breaking Changes

This removes the `ServeConfig.Listener` option we just introduced in 
the last patch release. In practice I think @paddycarver is the only 
consumer of this currently, so I propose just removing it while we're 
ahead.

If there are other consumers, I can revert that change. It was pretty
minor but I don't want to keep that functionality around if its not
necessary and the only reason we added it in the first place was
to facilitate this use case.

### DIFF
--- a/server.go
+++ b/server.go
@@ -419,17 +419,20 @@ func Serve(opts *ServeConfig) {
 		}
 	}
 
-	// Eat the interrupts
-	ch := make(chan os.Signal, 1)
-	signal.Notify(ch, os.Interrupt)
-	go func() {
-		count := 0
-		for {
-			<-ch
-			count++
-			logger.Trace("plugin received interrupt signal, ignoring", "count", count)
-		}
-	}()
+	// Eat the interrupts. In test mode we disable this so that go test
+	// can be cancelled properly.
+	if opts.Test == nil {
+		ch := make(chan os.Signal, 1)
+		signal.Notify(ch, os.Interrupt)
+		go func() {
+			count := 0
+			for {
+				<-ch
+				count++
+				logger.Trace("plugin received interrupt signal, ignoring", "count", count)
+			}
+		}()
+	}
 
 	// Set our stdout, stderr to the stdio stream that clients can retrieve
 	// using ClientConfig.SyncStdout/err.
--- a/server.go
+++ b/server.go
@@ -1,11 +1,13 @@
 package plugin
 
 import (
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"log"
 	"net"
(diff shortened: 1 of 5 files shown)


server.go
```diff
@@ -357,15 +395,29 @@ func Serve(opts *ServeConfig) {
...
+		// retry a few times.
+		ch <- &ReattachConfig{
+			Protocol: protoType,
+			Addr:     listener.Addr(),
+			Pid:      os.Getpid(),
+			Test:     true,
+		}
+	}
 
 	// Eat the interrupts
```
paddycarver, line 422: It'd be super cool if we could disable this in test mode; right now, tests can't be cancelled or killed by keyboard shortcut, you need to look up the process number and kill it with the `kill` unix command. Given the control of the process isn't supposed to reside with the client anymore, I would expect the process to respond to the interrupt signal.

server.go
```diff
@@ -379,18 +431,54 @@ func Serve(opts *ServeConfig) {
...
+	// using ClientConfig.SyncStdout/err.
+	//
+	// In test mode, we use a multiwriter so that the data continues going
+	// to the normal stdout/stderr so output can show up in test logs. We
+	// also send to the stdio stream so that clients can continue working
+	// if they depend on that.
+	if opts.Test != nil {
+		// In test mode we need to maintain the original values so we can
+		// reset it.
+		defer func(out, err *os.File) {
```
paddycarver, line 444: This is racy. Because os.Stdout is essentially global mutable state, if multiple servers are run in the same process in parallel (which happens in tests with some frequency), there's a good chance the following will happen:

* Server 1 sets os.Stdout to stdout_w, stores /dev/stdout in the defer
* Server 2 sets os.Stdout to stdout_w, stores stdout_w from Server 1 in the defer
* Server 1 finishes and the defer triggers, os.Stdout is now /dev/stdout
* Server 2 finishes and the defer triggers, os.Stdout is now stdout_w from Server 1

What's worse, once it happens, it will never be corrected; each new server will assume that os.Stdout is set to /dev/stdout and will reset it faithfully, so once a wrong value is persisted, it's persisted indefinitely. Even with a moderate number of tests, this race condition has a good chance of triggering. For large providers with lots of tests, it's almost certain to trigger.

I've pushed the branch `paddy-test-mode-race` with an update the test mode test to demonstrate this race condition.

server_test.go
```diff
@@ -1,12 +1,77 @@
...
+	}
+
+	// Canceling should cause an exit
+	cancel()
+	<-closeCh
+	if err := client.Ping(); err == nil {
+		t.Fatal("should error")
+	}
+
+	// Try logging, this should show out in tests. We have to manually verify.
```
paddycarver, line 71: I think at the start of the test we could persist `os.Stdout.Name()` and then compare it here. That wouldn't necessarily catch the race condition I noted, but it would catch other errors without relying on manual verification, I think.

server.go
mitchellh: Ah yes, this was designed on purpose assuming only one test was started at a time. Let's chat about your use case here and how multiple start at once.

server.go
mitchellh: Ah yep, good catch. Fixed.

server_test.go
mitchellh: We can't use `Name` at the moment because we overwrite it with a multiwriter, unless I'm missing something. :(

server.go
mitchellh: I think I have a solution to this, you now opt-in to stdio syncing with `SyncStdio` (in test config). This defaults to false since its messy/complex, which is the behavior you probably want.

paddycarver: (I know you fixed this, but to document it: the use case where multiple start at a time is when you use a server in multiple tests, and Go runs tests in parallel, there's no guarantee those tests won't run at the same time.)

server_test.go
paddycarver: In my test the multiwriter `Name()` returns `|1`, but I'm not sure this is important.

j0hnwhyte:
Hi and thanks for developing this. I was about to fork the repo to implement this by myself and just discovered you guys did it fairly recently, great job.
May I ask how one's supposed to integrate with the standard `gotest` command? It seems to avoid `main()`, so where should we initialize the plugin and tell the client to connect to us?

In theory the process should be:

* Tap the process startup when the program is built as a test
* Enable test mode on the `plugin.Serve()` call
* Wait for the `ReattachConfig` from the channel, pass it to the client via some custom protocol
* The client connects to us, we're able to prepare some context and services
* The test is actually ran, we fetch the context from (ugh..) global variables

Are there any comments about it? What do you guys think is the best approach?

paultyng:
@j0hnwhyte we implemented this in Terraform by using an environment variable to pass the "reattach" configuration for the plugin. Here are a bit more of the details: https://github.com/hashicorp/terraform/pull/24674

PR #160, opened 2020-10-19
Do not override the default log output in the server

### LINKED ISSUE #155: Standard log.Logger output overwritten
The standard log.Logger output gets overwritten when serving a plugin, and I'm not 100% clear on why. But it [creates a race condition](https://github.com/hashicorp/terraform-plugin-sdk/issues/543) in the Terraform Plugin SDK, which redirects the standard logger to a file or log level processor to avoid burying test results. Is this something we may be able to disable if `ServeConfig.Test` is not `nil`?

jbardin:
In order to transition large existing codebases to use end-to-end structured logging, we want to be able to make use of the hclog `Logger.StandardWriter`. The entails setting the default std library log output to a wrapped instance of an `hclog.Logger` while running the server.

The `plugin.Server` function however overrides the log output when initializing the server. This means the implementation currently needs to restore its logger again after the server has started (there is no callback, so this must be checked on the first method call). Failing to do so does not cause an error, however it does cause the log output to lose its structure. 

Removing the call to `log.SetOutput` makes it easier to setup a std library logger for all implementations, at the minor expense of requiring all implementations to ensure they write to `os.Stderr`, which is almost always the default case. 

Fixes #155


mitchellh (approved):
Yeah this seems good.

PR #163, opened 2020-11-23
Allow overriding ambient environment variables

### LINKED ISSUE #162: Can't override environment variables when running a plugin
I want to modify an existing environment variable _just for the plugin_ without affecting the environment variables of my current process. Right now any custom environment variables set on the plugin command are set first, and then the ambient environment variables are set second, undoing changes to an existing environment variable.

https://github.com/hashicorp/go-plugin/blob/0c19a1387af146eaa3335dad0a021741344ac3e9/client.go#L530

I'll follow-up with a PR to switch it to the following to enable this scenario if that's okay.

```go
cmd.Env = append(os.Environ(), cmd.Env...) 
```

carolynvs:
Change when os.Environ() is applied so that the environment variables explicitly set on the plugin command take precedence over ambient environment variables.

Fixes #162

### DIFF
--- a/client.go
+++ b/client.go
@@ -524,7 +524,7 @@ func (c *Client) Start() (addr net.Addr, err error) {
 	}
 
 	cmd := c.config.Cmd
-	cmd.Env = append(cmd.Env, os.Environ()...)
+	cmd.Env = append(os.Environ(), cmd.Env...)
 	cmd.Env = append(cmd.Env, env...)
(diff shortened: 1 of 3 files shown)


carolynvs:
I have signed the CLA

ernestrc:
Can we get this merged?

swenson (approved):
LGTM

PR #171, opened 2021-03-31
add ProtocolVersion to ReattachConfig

mildwonkey:
This PR adds the (already-determined) `ProtocolVersion` to `ReattachConfig`. This information was not captured previously, so a caller connecting to a running process had no way of knowing what `ProtocolVersion` had been negotiated:  `client.NegotiatedVersion` would always return `0`.


jbardin (approved):
LGTM

PR #173, opened 2021-08-10
fix handles leakage

a-zagaevskiy:
The leakage was detected during using of Nomad v0.10.4 on Windows 10: https://github.com/hashicorp/nomad/issues/11026

### DIFF
--- a/process_windows.go
+++ b/process_windows.go
@@ -19,6 +19,7 @@ func _pidAlive(pid int) bool {
 	if err != nil {
 		return false
 	}
+	defer syscall.CloseHandle(h)
 
 	var ec uint32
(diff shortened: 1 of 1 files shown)


notnoop (approved):
Thanks @a-zagaevskiy !

For future readers, Windows API docs states that process handlers should be released:

> When you are finished with the handle, be sure to close it using the [CloseHandle](https://docs.microsoft.com/en-us/windows/win32/api/handleapi/nf-handleapi-closehandle) function.

https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-openprocess

PR #186, opened 2022-01-04
Trim a spurious log message for plugins using RPC

### LINKED ISSUE #72: Basic example seems to be broken (at least on Mac)
I the main program and here is what I'm getting:

```
bash-3.2$ ./basic
2018-07-05T00:32:05.780+0500 [DEBUG] plugin: starting plugin: path=./plugin/greeter args=[./plugin/greeter]
2018-07-05T00:32:05.784+0500 [DEBUG] plugin: waiting for RPC address: path=./plugin/greeter
2018-07-05T00:32:05.870+0500 [DEBUG] plugin.greeter: message from plugin: foo=bar timestamp=2018-07-05T00:32:05.868+0500
2018-07-05T00:32:05.871+0500 [DEBUG] plugin.greeter: plugin address: address=/var/folders/4v/3yrsz17s1m9bqlj35bgqvyvr0000gn/T/plugin847884065 network=unix timestamp=2018-07-05T00:32:05.871+0500
2018-07-05T00:32:05.877+0500 [DEBUG] plugin.greeter: message from GreeterHello.Greet: timestamp=2018-07-05T00:32:05.876+0500
Hello!
2018-07-05T00:32:05.877+0500 [DEBUG] plugin.greeter: 2018/07/05 00:32:05 [ERR] plugin: plugin server: accept unix /var/folders/4v/3yrsz17s1m9bqlj35bgqvyvr0000gn/T/plugin847884065: use of closed network connection
2018-07-05T00:32:05.878+0500 [DEBUG] plugin: plugin process exited: path=./plugin/greeter
```

Is "use of closed network connection" expected here? I tried running bidirectional example and there is no problem.

tbehling:
This PR trims a spurious log message for plugins using RPC.

When a plugin is shutting down, the plugin host's log will often report an error about `accept unix` / `use of closed network connection` .  Essentially, an `accept` call is getting invoked on a socket that is already closed.  This PR simply updates the RPC server's `accept` loop to avoid logging that error.

This PR makes an assumption that there are no legitimate debugging situations where this error indication is helpful.

As reported in #72, when using the "basic" example in this repo, the output is the following:

```
2022-01-04T15:36:16.041-0500 [DEBUG] plugin: starting plugin: path=./plugin/greeter args=[./plugin/greeter]
2022-01-04T15:36:16.041-0500 [DEBUG] plugin: plugin started: path=./plugin/greeter pid=10674
2022-01-04T15:36:16.041-0500 [DEBUG] plugin: waiting for RPC address: path=./plugin/greeter
2022-01-04T15:36:16.046-0500 [DEBUG] plugin.greeter: message from plugin: foo=bar timestamp=2022-01-04T15:36:16.046-0500
2022-01-04T15:36:16.046-0500 [DEBUG] plugin.greeter: plugin address: address=/tmp/plugin3230253060 network=unix timestamp=2022-01-04T15:36:16.046-0500
2022-01-04T15:36:16.046-0500 [DEBUG] plugin: using plugin: version=1
2022-01-04T15:36:16.047-0500 [DEBUG] plugin.greeter: message from GreeterHello.Greet: timestamp=2022-01-04T15:36:16.047-0500
Hello!
2022-01-04T15:36:16.047-0500 [DEBUG] plugin.greeter: 2022/01/04 15:36:16 [ERR] plugin: plugin server: accept unix /tmp/plugin3230253060: use of closed network connection
2022-01-04T15:36:16.047-0500 [DEBUG] plugin: plugin process exited: path=./plugin/greeter pid=10674
2022-01-04T15:36:16.047-0500 [DEBUG] plugin: plugin exited
```

With this patch, the line about `use of closed network connection` is absent.

Fixes #72 
Fixes https://github.com/cloud-native-go/examples/issues/2

### DIFF
--- a/client_posix_test.go
+++ b/client_posix_test.go
@@ -1,3 +1,4 @@
+//go:build !windows
 // +build !windows
 
 package plugin
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/hashicorp/go-plugin
 
-go 1.13
+go 1.17
 
 require (
 	github.com/golang/protobuf v1.3.4
@@ -9,7 +9,16 @@ require (
 	github.com/jhump/protoreflect v1.6.0
 	github.com/mitchellh/go-testing-interface v0.0.0-20171004221916-a61a99592b77
 	github.com/oklog/run v1.0.0
-	github.com/stretchr/testify v1.3.0 // indirect
 	golang.org/x/net v0.0.0-20190311183353-d8887717615a
 	google.golang.org/grpc v1.27.1
+)
+
+require (
+	github.com/fatih/color v1.7.0 // indirect
+	github.com/mattn/go-colorable v0.1.4 // indirect
+	github.com/mattn/go-isatty v0.0.10 // indirect
+	github.com/stretchr/testify v1.3.0 // indirect
+	golang.org/x/sys v0.0.0-20191008105621-543471e840be // indirect
+	golang.org/x/text v0.3.0 // indirect
+	google.golang.org/genproto v0.0.0-20190819201941-24fa4b261c55 // indirect
 )
--- a/go.sum
+++ b/go.sum
@@ -2,7 +2,6 @@ cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMT
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
 github.com/census-instrumentation/opencensus-proto v0.2.1/go.mod h1:f6KPmirojxKA12rnyqOA5BBL4O983OfeGPqjHWSTneU=
 github.com/client9/misspell v0.3.4/go.mod h1:qj6jICC3Q7zFZvVWo7KLAzC3yx5G7kyvSDkc90ppPyw=
-github.com/davecgh/go-spew v1.1.0 h1:ZDRjVQ15GmhC3fiQ8ni8+OwkZQO4DARzQgrnXU1Liz8=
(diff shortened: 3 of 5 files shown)


jbardin:
Thanks @tbehling! I think it may still be useful to log the closed case as debug, so that there's an indication why the server exited if the caller wasn't expecting it. 

The `ErrClosed` value was only recently added, so we will need to bump the go version for the module too.

(also make sure to run `go fmt`, there's an extra space in the if statement)

tbehling:
Hi @jbardin, thank you so much for giving this a review, and for our subsequent conversation.

Based on your comments, I've updated this PR to retain the log message, but print it with `[DEBUG]` instead of `[ERR]`.  That should eliminate the pain of seeing false-positive errors in logs for these benign messages.

I see the `go.mod` here specifies Go 1.13.  `net.ErrClosed` was added in Go 1.16 (and changed data types in 1.17!).  The code here uses some Go tricks to ensure backward compatibility, while still using `errors.Is()`.

I don't love the complexity, but the alternative is to compare the Go error text, which has been done by a lot of Go projects in the past, but is fragile, and the text has changed before.

tbehling:
Hi again @jbardin, per our discussion, I have dropped the compatibility with old versions of Go, which simplifies this PR a lot.

If I now build the `examples/basic` plugin, the relevant line from the output of `./basic` is:

`2022-01-27T16:28:25.716-0500 [DEBUG] plugin.greeter: 2022/01/27 16:28:25 [DEBUG] plugin: plugin server: accept unix /tmp/plugin3304580809: use of closed network connection`

The difference from the original is that second `[DEBUG]` used to be `[ERR]`.

I also bumped `go.mod` to Go 1.17, and ran `go fmt` which made updates to 2 files.

This PR doesn't have Go CI tests, but locally, `go test` passes.

jbardin (approved):

tbehling:
For reference there is some discussion about Go's `net.ErrClosed` at https://github.com/golang/go/issues/4373 .

PR #195, opened 2022-05-03
increase level of plugin exit logs

gavriel-hc:
These logs are pretty valuable, they shouldn't just be at a DEBUG level

### DIFF
--- a/client.go
+++ b/client.go
@@ -638,7 +638,7 @@ func (c *Client) Start() (addr net.Addr, err error) {
 				[]interface{}{"error", err.Error()}...)
 			c.logger.Error("plugin process exited", msgArgs...)
 		} else {
-			// Log and make sure to flush the logs write away
+			// Log and make sure to flush the logs right away
 			c.logger.Info("plugin process exited", msgArgs...)
 		}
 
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -629,17 +629,19 @@ func (c *Client) Start() (addr net.Addr, err error) {
...
 			"path", path,
 			"pid", pid,
 		}
 		if err != nil {
-			debugMsgArgs = append(debugMsgArgs,
+			msgArgs = append(msgArgs,
 				[]interface{}{"error", err.Error()}...)
+			c.logger.Error("plugin process exited", msgArgs...)
+		} else {
+			// Log and make sure to flush the logs write away
```
tbehling, line 641:
```suggestion
			// Log and make sure to flush the logs right away
```

tbehling:
For the record, I had verbally LGTM'ed this PR while pairing with @gavriel-hc.

kaovilai:
Opt out of info plugin exit logs

We may want ability to opt into silencing these logs per https://github.com/vmware-tanzu/velero/issues/8653#issuecomment-2621451900

PR #219, opened 2022-10-19
fix: do not kill plugin process when use examples to test

diannaowa:
Do not kill plugin process when use "examples" to test.

501 95990     1   0  5:19PM ttys023    0:00.01 ./kv-go-grpc
501 96014     1   0  5:19PM ttys023    0:00.01 ./kv-go-grpc
501 96038     1   0  5:19PM ttys023    0:00.01 ./kv-go-grpc
501 96064     1   0  5:19PM ttys023    0:00.01 ./kv-go-grpc
501 96088     1   0  5:19PM ttys023    0:00.01 ./kv-go-grpc
501 96112     1   0  5:19PM ttys023    0:00.01 ./kv-go-grpc
501 96138     1   0  5:19PM ttys023    0:00.01 ./kv-go-grpc
501 96166     1   0  5:19PM ttys023    0:00.01 ./kv-go-grpc
501 96190     1   0  5:19PM ttys023    0:00.01 ./kv-go-grpc
501 97047 95287   0  5:43PM ttys023    0:00.00 grep grpc

### DIFF
--- a/examples/bidirectional/main.go
+++ b/examples/bidirectional/main.go
@@ -36,14 +36,14 @@ func main() {
 	rpcClient, err := client.Client()
 	if err != nil {
 		fmt.Println("Error:", err.Error())
-		os.Exit(1)
+		return
 	}
 
 	// Request the plugin
 	raw, err := rpcClient.Dispense("counter")
 	if err != nil {
 		fmt.Println("Error:", err.Error())
-		os.Exit(1)
+		return
 	}
 
 	// We should have a Counter store now! This feels like a normal interface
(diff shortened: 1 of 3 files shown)


examples/bidirectional/main.go
```diff
@@ -36,14 +36,14 @@ func main() {
 	rpcClient, err := client.Client()
 	if err != nil {
 		fmt.Println("Error:", err.Error())
-		os.Exit(1)
+		return
```
swenson, line 39: What is the logic behind switching to `return`? We lose the error code when we don't use `os.Exit()`.

That said, this is only example code.

PR #223, opened 2022-12-02
Add more helpful message on plugin setup failure

swenson:
The current error message shown when a plugin does not respond correctly can be obtuse to the end user of, for example, Vault.

Here, we add potential reasons why the plugin did not respond correctly, and some additional debugging information (provided as a best effort) including the CPU architecture that the plugin was compiled for, the current CPU architecture, and the permissions of the plugin. Hopefully this will help users diagnose why their plugin is not loading.

We also added a `testdata/` directory that is optionally populated with executables in various formats to help test the additional debugging information. The binaries are over 1 MB each though, so are not checked in, and the test will be skipped if they have not been compiled.

The error message comes out looking like:

```
    err: Unrecognized remote plugin message: something invalid here
        This usually means
          the plugin was not compiled for this architecture,
          the plugin is missing dynamic-link libraries necessary to run,
          the plugin is not readable by this process, or
          the plugin was compiled with an incompatible version of the go-plugin protocol

        Additional notes about plugin:  Path: /var/folders/jb/4vzywqts2gl49jvwp0pz9dsc0000gq/T/go-build2701257428/b001/go-plugin.test
          Mode: -rwxr-xr-x
          MachO architecture: CpuArm64 (current architecture: arm64)
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -34,7 +34,7 @@ const unrecognizedRemotePluginMessage = `Unrecognized remote plugin message: %s
 This usually means
   the plugin was not compiled for this architecture,
   the plugin is missing dynamic-link libraries necessary to run,
-  the plugin is not readable by this process, or
+  the plugin is not executable by this process due to file permissions, or
   the plugin failed to negotiate the initial go-plugin protocol handshake
 %s`
 
--- a/client.go
+++ b/client.go
@@ -6,6 +6,9 @@ import (
 	"crypto/subtle"
 	"crypto/tls"
 	"crypto/x509"
+	"debug/elf"
+	"debug/macho"
+	"debug/pe"
 	"encoding/base64"
 	"errors"
 	"fmt"
@@ -16,6 +19,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 	"sync"
@@ -25,6 +29,14 @@ import (
 	"github.com/hashicorp/go-hclog"
 	"google.golang.org/grpc"
 )
+
+const unrecognizedRemotePluginMessage = `Unrecognized remote plugin message: %s
+This usually means
+  the plugin was not compiled for this architecture,
+  the plugin is missing dynamic-link libraries necessary to run,
+  the plugin is not readable by this process, or
+  the plugin was compiled with an incompatible version of the go-plugin protocol
+%s`
 
 // If this is 1, then we've called CleanupClients. This can be used
 // by plugin RPC implementations to change error behavior since you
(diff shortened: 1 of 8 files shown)


client.go
```diff
@@ -473,7 +485,43 @@ func (c *Client) Kill() {
...
+	} else if machoFile, err := macho.Open(path); err == nil {
+		notes += fmt.Sprintf("  MachO architecture: %s (current architecture: %s)\n", machoFile.Cpu, runtime.GOARCH)
+	} else if peFile, err := pe.Open(path); err == nil {
+		machine, ok := peTypes[peFile.Machine]
+		if !ok {
+			machine = "unknown"
+		}
+		notes += fmt.Sprintf("  PE architecture: %s (current architecture: %s)\n", machine, runtime.GOARCH)
+	}
+	return
```
tomcf-hcp, line 521: `return notes` to avoid naked return?

client.go
```diff
@@ -473,7 +485,43 @@ func (c *Client) Kill() {
...
+	0x8664: "amd64",
+	0xaa64: "arm64",
+}
+
+// additionalNotesAboutCommand tries to get additional information about a command that might help diagnose
+// why it won't run correctly. It runs as a best effort only.
+func additionalNotesAboutCommand(path string) (notes string) {
+	stat, err := os.Stat(path)
+	if err != nil {
+		return
```
tomcf-hcp, line 503: `return ""` or something other than a naked return here?

client.go
swenson: It's not necessary since `notes` a named return parameter. (And a naked return is the style I usually see for these things.)

client.go
```diff
@@ -473,7 +485,43 @@ func (c *Client) Kill() {
...
+	}
+
+	notes += "\nAdditional notes about plugin:"
```
calvn, line 506:
```suggestion
	notes += "\nAdditional notes about plugin:\n"
```

tomhjp (commented):
This is great 👍 left a few thoughts for minor improvements.

client.go
```diff
@@ -26,6 +30,14 @@ import (
 	"google.golang.org/grpc"
 )
 
+const unrecognizedRemotePluginMessage = `Unrecognized remote plugin message: %s
+This usually means
+  the plugin was not compiled for this architecture,
+  the plugin is missing dynamic-link libraries necessary to run,
+  the plugin is not readable by this process, or
```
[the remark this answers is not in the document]
swenson: I thought about this, but the file doesn't necessarily need to have the `+x` bit set, just be accessible, I believe.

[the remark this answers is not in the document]
tomhjp: How come we don't need `x`? We execute the binary here don't we? (Sorry if this is a stupid question)

https://github.com/hashicorp/go-plugin/blob/b60690f8999f79dae652a67e7b81a6923b43cff8/client.go#L586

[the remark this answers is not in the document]
swenson: Oh, sorry, you're right. I was thinking about dynamic linked libraries, which don't need to be executable. :)

client.go
```diff
@@ -473,7 +485,44 @@ func (c *Client) Kill() {
...
+func additionalNotesAboutCommand(path string) string {
+	notes := ""
+	stat, err := os.Stat(path)
+	if err != nil {
+		return notes
+	}
+
+	notes += "\nAdditional notes about plugin:\n"
+	notes += fmt.Sprintf("  Path: %s\n", path)
+	notes += fmt.Sprintf("  Mode: %s\n", stat.Mode())
```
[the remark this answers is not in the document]
swenson: Added:

```
err: Unrecognized remote plugin message:
        This usually means
          the plugin was not compiled for this architecture,
          the plugin is missing dynamic-link libraries necessary to run,
          the plugin is not readable by this process, or
          the plugin was compiled with an incompatible version of the go-plugin protocol

        Additional notes about plugin:
          Path: /var/folders/jb/4vzywqts2gl49jvwp0pz9dsc0000gq/T/go-build2608285047/b001/go-plugin.test
          Mode: -rwxr-xr-x
          Owner: 503 [swenson] (current: 503 [swenson])
          Group: 20 [staff] (current: 20 [staff])
```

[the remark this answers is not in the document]
tomhjp: Nice!

client.go
```diff
@@ -697,10 +769,7 @@ func (c *Client) Start() (addr net.Addr, err error) {
 		line = strings.TrimSpace(line)
 		parts := strings.SplitN(line, "|", 6)
 		if len(parts) < 4 {
-			err = fmt.Errorf(
-				"Unrecognized remote plugin message: %s\n\n"+
-					"This usually means that the plugin is either invalid or simply\n"+
-					"needs to be recompiled to support the latest protocol.", line)
+			err = fmt.Errorf(unrecognizedRemotePluginMessage, line, additionalNotesAboutCommand(cmd.Path))
```
tomhjp, line 772: Open question: Given the scope of usage of this library, and the fact we have a levelled logger available, would it potentially make sense to log the additional notes to debug level instead, and include a note about the availability of debug logs in the error message? I can definitely see the appeal of just having all the information available right there in the error message, and I can also see the possibility for grumbles about verboseness (which I personally don't sympathise with too much), so I'm genuinely undecided.

tomhjp (approved):
I left a question but I am happy with either direction, so LGTM 👍

client.go
swenson: I think by returning an error message, we give the caller the option to log and at what level. They certainly might perform an action based on the remote plugin message in the first line. I think I prefer it as is.

fairclothjm (approved):
👍

swenson:
Thanks!

client.go
```diff
@@ -473,7 +487,65 @@ func (c *Client) Kill() {
...
+	notes := ""
+	stat, err := os.Stat(path)
+	if err != nil {
+		return notes
+	}
+
+	notes += "\nAdditional notes about plugin:\n"
+	notes += fmt.Sprintf("  Path: %s\n", path)
+	notes += fmt.Sprintf("  Mode: %s\n", stat.Mode())
+	statT, ok := stat.Sys().(*syscall.Stat_t)
```
drakkan, line 512: Hi,

this does not compile on Windows:

```
Error: C:\Users\runneradmin\go\pkg\mod\github.com\hashicorp\go-plugin@v1.4.7\client.go:512:36: undefined: syscall.Stat_t
```

https://github.com/drakkan/sftpgo/actions/runs/3641265803/jobs/6146996350

Issue #213, opened 2022-09-24
Use custom listener when serving


Anthony-Jhoiro:
I really enjoy using this package, but I encountered a limitation when trying to serve plugins that listen on TCP instead than with UNIX sockets due to security limitations.

The `Serve` function, as shown in the examples, seems only to create UNIX sockets listeners when running on non-windows system.

I tried a few things, first with a manually created server like this : 
```go
listener, err := net.Listen("tcp", ":12345")
if err != nil { panic(err) }

server := plugin.GRPCServer{
     // ...plugin config
}

err = server.Init()
if err != nil { panic(err) }

server.Serve(listener)
```

But I encountered an issue with the logger, as it is not initialized in the `server.Init` function, so the program crashes anytime, it tries to log with a segmentation error.

I made it work by adding a field for a custom listener in the `ServeConfig` struct (server.go:58) and preventing calling the `serverListener()` function in the `Serve` function. This way I can use the TCP listener like this : 
```go
listener, err := net.Listen("tcp", ":12345")
if err != nil {
    panic(err)
}

plugin.Serve(&plugin.ServeConfig{
    // ...plugin config
    Listener:   listener,
})
```

And this seems to be working fine.

I am not sure to fully understand this library that's why I didn't make a PR about this now. Is there another planned way of doing that? I can do a PR about it if needed. 
If it helps, you can find my fork here : [https://github.com/Anthony-Jhoiro/go-plugin/tree/feat/serve_custom_listener](https://github.com/Anthony-Jhoiro/go-plugin/tree/feat/serve_custom_listener)

I am using the latest package version and go 1.18.3 btw.
Thanks a lot !

davestibrany-dd:
+1 this would be very useful, or even just providing us with a way to configure tcp/unix sockets in ServeConfig.

Issue #151, opened 2020-07-14
permission denied on sharing unix domain sockets with across different user


msays2000:
Mother host process running as user A. Plugin process running with sudo i.e root. 

Plugin creates a unix domain socket at /tmp/plugin226668393 but mother host is unable to connect. 

`rpc error: code = Unavailable desc = connection error: desc = "transport: error while dialing: dial unix /tmp/plugin226668393: connect: permission denied"
`
Possible solutions:
- Provide options to create socket in a common shareable location (/var/go-plugin). Something like under /var/go-plugin/pluginxxxxxxx 
Reference: https://github.com/hashicorp/go-plugin/blob/master/server.go#L547-L574
Tutorial: https://medium.com/@aeronanubhav/sharing-unix-domain-sockets-with-multiple-users-4c7b4d439c39

- Provide options to use tcp instead of unix domain socket in linux. https://github.com/hashicorp/go-plugin/blob/master/server.go#L497-L503

tomhjp:
This will be because the plugin's UDS listener is owned by root with 0o600 permissions. The library recently got a new [ClientConfig.UnixSocketConfig](https://github.com/hashicorp/go-plugin/blob/303d84fc850fc2ad18981220339702809f8be06a/client.go#L246) option which offers a solution to this situation. You can set `Group` to any group that user A is in, and it will change the group owner of the socket to that group, and set the permissions on the socket to 0o660 so that user A gets rw permission to the socket via its group membership.

Note that the plugin's group membership (and whether it's root or not) doesn't matter because it retains rw permission through the user owner and permission bits, just like it did previously.

PR #226, opened 2022-12-07
allow network type to be specified

davestibrany-dd:
Allows users to specify a network type of "tcp" or "unix. If no network type is specified, we'll default to "tcp" for windows, and "unix" otherwise.

Related issues: 
#213 
#151

### DIFF
--- a/server.go
+++ b/server.go
@@ -85,6 +85,10 @@ type ServeConfig struct {
 	// server will create a default logger.
 	Logger hclog.Logger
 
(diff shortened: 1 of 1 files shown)


davestibrany-dd:
For our particular use case, we've noticed high kernel memory usage when using unix sockets. We're continuously spinning up/tearing down short lived plugins which leads to bloat in the linux dentry cache. Even though this memory is reclaimable by the kernel, it makes it difficult to monitor our containers for legit memory leaks. Switching to tcp sockets fixes this for us.

PR #240, opened 2023-03-03
deps: Remove direct dependency on golang.org/x/net

bflad:
Reference: https://pkg.go.dev/golang.org/x/net@v0.0.0-20190311183353-d8887717615a/context
Reference: https://github.com/hashicorp/go-plugin/pull/186

Previously, the module was only using `golang.org/x/net/context` import. That package was deprecated with Go 1.7 when its functionality migrated into the standard library `context` package. This Go module already announces Go 1.17 minimum since version 1.4.4 and the `context` types are not exposed as part of any exported API.

### DIFF
--- a/examples/bidirectional/proto/kv.pb.go
+++ b/examples/bidirectional/proto/kv.pb.go
@@ -22,7 +22,7 @@ import fmt "fmt"
 import math "math"
 
 import (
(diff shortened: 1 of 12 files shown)


fairclothjm (approved):
Thanks!

christophermancini (approved):

PR #241, opened 2023-03-04
additional notes: ensure to close files

### LINKED ISSUE #232: Files are not closed on windows
[additionalNotesAboutCommand](https://github.com/hashicorp/go-plugin/blob/main/notes_windows.go#L17-L39) is opening a file using `elf`, `macho` and `pe`  to extract information,
but those files are never closed. 
This can cause issues when another process is trying to open them again or try to remove.

drakkan:
Fixes #232


fairclothjm (approved):
Thanks!

PR #242, opened 2023-03-06
update deps

drakkan:
fixes x/net and x/text security issues


tomhjp:
Sorry this went unanswered for so long, but I merged #285 with more recent deps a couple of weeks ago, so I think that covers this PR for now. It hasn't been released in a tag yet, but I'll be cutting one in the next few weeks or so, and you can force the version used in consuming applications before then via Go's MVS.

PR #250, opened 2023-04-11
Fix gRPC and bidirectional examples' code generation instructions

### LINKED ISSUE #249: gRPC example's proto regeneration command doesn't work
Hi there, forgive me for any ignorance, as I'm a total newb with gRPC, but I tried the gRPC example to see if it would work (specifically, if generating Golang code from the provided `.proto` file would result in the same or similar results as what's already checked in to this repo) because I'm planning to use this repo's plugin system in my company's project and I want to make sure that I can reference the examples to make sure I'm on the right track. However, when I follow [the instructions from the README](https://github.com/hashicorp/go-plugin/tree/main/examples/grpc#updating-the-protocol), I get the following error:

```
% protoc -I proto/ proto/kv.proto --go_out=plugins=grpc:proto/
protoc-gen-go: unable to determine Go import path for "kv.proto"

Please specify either:
        • a "go_package" option in the .proto source file, or
        • a "M" argument on the command line.

See https://developers.google.com/protocol-buffers/docs/reference/go-generated#package for more information.

--go_out: protoc-gen-go: Plugin failed with status code 1.
```

Here's my version of `protoc` in case it matters:

```
% protoc --version
libprotoc 3.15.8
```

I did notice that the README hasn't been updated in 6 years, so my guess is that either there's something that changed with `protoc` since then, or maybe there's some unspoken assumption about the development environment that I'm supposed to have met before I started on this, but I did check to make sure that the example compiled and worked as-is without changing anything about my development environment, and since it does, I _feel like_ it's not PEBKAC in this case. Regardless, it would be nice to get an update to the instructions either way: if it's a matter of something about `protoc` changing in the past 6 years, then the code/instructions should change accordingly, and if it's a matter of me just not meeting some base assumptions about the development environment, then having those assumptions mentioned, if not spelled out, would be really useful. It would probably also be good to mention which version of `protoc` was used to generate the Golang code.

Thank you for taking the time to read this!

2rs2ts:
Hello! My teammates and I were trying to get the `grpc` example to work, and were eventually able to figure out the things that needed to change to make its instructions, which seem to be around six years old, work with modern versions of `protoc` and the related Golang tooling for it. We decided that not only should we update the `.proto` file and generated code, we should also make sure that the two separate generated files (since there are two now, rather than one) get committed. In addition to that, we recorded some details in the README about how to find the versions of the software used to generate the code, for newbies to the codebase (like ourselves!) who didn't really know what was assumed about the development environment.

The updates to the client required us to also regenerate the Python plugin code, so we updated that and made sure we spelled out the dependencies that were used for that language. We found that we did have to manually re-insert the copyright notices to the python code, but not the Golang code; if there's an option in the generation command that fixes that, then we'd love to know about it.

We felt that recording the versions of various software dependencies used would not only help orient new users, but it would also make it easier in the future, if `protoc` or related tools introduce more breaking changes, to figure out what needs to change to make the instructions keep working. This is our first time using both this plugin repo and gRPC in general, and we're not experts with Golang either, so it may read as unnecessarily redundant information to an experienced developer, so we apologize for that. Also, we may not have written idiomatic code, but we're happy to take feedback and make sure it is idiomatic.

While we were at it, we also made the `net/rpc` example that lived in the `grpc` example's folder work. Before, it would fail with an error from `rpc_server.go` saying that it couldn't find the (hardcoded) `kv_grpc` plugin, which makes sense since the `net/rpc` plugin doesn't define that for obvious reasons.

And we also updated the `bidirectional` example since it's similar to the `grpc` example, plus we tested the other examples to ensure they still work, particularly the `negotiated` example which re-uses the `grpc` example's code. They all work on our machine, so hopefully they all work for you too.

Now, probably the most important change here is not to the example, but rather the top-level `go.mod` dependencies. The thing is, we were struggling to get things to work unless we upgraded the gRPC and protobuf dependencies to match modern guides (such as [this one on `grpc.io`](https://grpc.io/docs/languages/go/quickstart)) unless we updated those dependencies. We were really hesitant to proceed until we noticed that [the `terraform-plugin-go` repo uses the same `grpc` version and even a much more recent `protobuf` version,](https://github.com/hashicorp/terraform-plugin-go/blob/0ad974e4833d52dfea55c1384d06ce678ed748d8/go.mod#L14-L16) and if that repo is supposed to serve as a template or guide for people writing terraform plugins, and it's pretty married to gRPC, then upgrading would probably be fine. Also, our new generation instructions are based not only on those modern guides, but also how [that repo generates its code too](https://github.com/hashicorp/terraform-plugin-go/blob/0ad974e4833d52dfea55c1384d06ce678ed748d8/Makefile#L6-L25), so we felt pretty justified in our decisions after that. With that said, we understand that this repo is a dependency for many Hashicorp projects, so we're not sure whether we actually _can_ upgrade those dependencies. Please let us know if we need to find a set of dependencies that will work for all consumers of this project.

At any rate, this fixes #249.


2rs2ts:
Please accept my apologies about the delay in signing the CLA. I did not realize I had not already signed the CLA a long time ago, and I'm checking with my employer's legal department to make sure it's ok for me to sign it.

2rs2ts:
I got approval to sign from my employer. I am surprised that the bot does not require the commit co-authors to also sign, but I'll ask them to do it too.

heaxo:
Unknown flag: --go-grpc_opt

2rs2ts:
@heaxo you might have an out of date version of `protoc` or the golang/grpc plugins for it. While I used slightly newer versions of the libraries from this guide than are printed here (as you can tell from the generated code), I do believe the versions from here will also work: https://grpc.io/docs/languages/go/quickstart/

2rs2ts:
@tomhjp is my PR obsoleted by https://github.com/hashicorp/go-plugin/pull/286? I haven't had time to look deeply into it, but I'd be happy to close my PR if you can confirm

tomhjp:
Hey @2rs2ts - I think you're right that most of this is obsoleted by #286, sorry about that and the delay responding. Comparing against main, it does look like your fix for the `net/rpc` example within `grpc` could still be landed, as it still looks broken to me on main. If you did want to land it, you could rebase the PR down to that change, but unfortunately I won't be help land that as I'll be losing write access to this repo later today. Thank you for the PR though! ❤️

2rs2ts:
It's been too long, and I don't remember a whole lot, but it seems that the new generation process would necessitate me starting from scratch, and I'd have to rewrite the PR summary. At that point, why even bother saving this branch? Hopefully this PR will be of some use to someone who does want to re-update the example for `net/rpc`. (We were not using it, we just made sure that if we were going to change one thing, we'd keep it all consistent...)

PR #257, opened 2023-06-05
GRPCBroker: add DialWithOptions

jmank88:
When creating new clients and servers with a `type GRPCBroker`, `grpc.ServerOption`s can be added when using `Accept(id uint32)` and `AcceptAndServe(id uint32, s func([]grpc.ServerOption) *grpc.Server)`, but `grpc.DialOption`s cannot be added when using `Dial(id uint32)`.

This PR proposes adding a `DialWithOptions(id uint32, opts ...grpc.DialOption)` method.

Note: `Dial` could be updated in place without affecting typical uses, but the signature change would still be breaking.
```diff
- Dial(id uint32)
+ Dial(id uint32, opts ...grpc.DialOption)
``` 
(this is the approach proposed in #251)

### DIFF
--- a/grpc_broker.go
+++ b/grpc_broker.go
@@ -523,9 +523,12 @@ func (b *GRPCBroker) muxDial(id uint32) func(string, time.Duration) (net.Conn, e
 }
 
 // Dial opens a connection by ID.
-func (b *GRPCBroker) Dial(id uint32) (conn *grpc.ClientConn, err error) {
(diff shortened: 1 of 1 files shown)


jmank88:
Rebased muxer conflicts.

jmank88:
@tomhjp @VioletHynes Is someone available to consider this feature (or #251)? :pray:

VioletHynes:
Hey! This seems super reasonable. Thanks for raising this. I'll try and get somebody's eyes on this shortly.

divyaac (approved):

PR #262, opened 2023-07-14
fix(examples): move client logic to a function that always return

willyrgf:
This commit fixes a bug in the `main()` that runs `defer client.Kill()` but use `os.Exit()` to finish the program; defer is just executed when a function returns. Because of that, the plugin was never closed. 

To fix it, I moved the client logic to a separate function that always returns and let the `os.Exit()` for the `main()`. It will guarantee that we will always call the `defer client.Kill()`.


tomhjp (approved):
LGTM, thanks for the fix!

PR #264, opened 2023-07-24
deps: bump protoreflect to v1.15.1

vancluever:
This bumps protoreflect to v1.15.1, which should correct some long-running issues WRT downstream dependencies and protoreflect's use of grpc_testing, unblocking downstream dependencies from upgrading the grpc package again.


johanbrandhorst (commented):
This upgrades to the v2 protobuf API, will this have any breaking change consequences on downstream packages?

vancluever:
@johanbrandhorst hard for me to say.

My fork is working in https://github.com/vancluever/terraform-provider-acme/commit/50a0f940f5e5ff53602eab9d935e54b7f4e60b75 and forward for passing tests (see https://github.com/vancluever/terraform-provider-acme/tree/64764c71c1188bb1c461112b021c49e272ac4c18), and my tests are working through a go-plugin implementation for DNS  providers. For other things in the go-plugin ecosystem though, I don't currently have visibility on, unfortunately.

johanbrandhorst (approved):
This is a nice step towards upgrading to the v2 API wholesale, but the consequences for the ecosystem is hard to predict. Maybe we can make a release candidate with this change before making a full release, to gauge any unintended impact?

vancluever:
@johanbrandhorst works for me! The ACME plugin is a pretty good isolated case and given my own testing, I think that combined with its stability within Terraform itself would be a pretty good test of how the larger ecosystem would fare.

PR #265, opened 2023-07-26
Allow for long log lines

luoxiaohei:
This pull request introduces a new feature to the `ClientConfig` in the `client.go`. The main change involves the addition of a `LogBufferSize` field, which will serve as the buffer size for reading stderr logs.

Currently, the `stdErrBufferSize` is set to a default value of 64 KB. When a log line exceeds this size, the client will not perform json unmarshal and subsequent processing for that line. This behavior can be unfriendly for certain use cases that involve lengthy logs.

To address this limitation, the proposed solution is to add the `LogBufferSize` field to the `ClientConfig`, allowing users to customize their own buffer size when creating a client. If the user doesn't specify a value, the default of 64 KB will be used to maintain compatibility with the current implementation.

An alternative approach would be to handle oversized logs by creating a new buffer to retain truncated data, concatenating it with historical data on the next read, and then proceeding with the complete log unmarshal. However, this method introduces added complexity and could lead to performance overhead in scenarios where logs consistently exceed the buffer size.

Considering simplicity and ease of implementation, the configuration-based solution is preferred. Users can now effortlessly fine-tune the buffer size to suit their specific logging needs.


tomhjp:
Hey @luoxiaohei - my plate is pretty full atm, but just wanted to say this looks like something we should be able to land and I'll try to revisit it soon. Thanks for putting this together, I can see it's been very thoughtfully constructed.

luoxiaohei:
> Hey @luoxiaohei - my plate is pretty full atm, but just wanted to say this looks like something we should be able to land and I'll try to revisit it soon. Thanks for putting this together, I can see it's been very thoughtfully constructed.

Thanks for your positive feedback on my pr, I look forward to seeing it merged soon.

tomhjp:
Thinking out loud a bit, an alternative solution would be to allow lines that cross the buffer threshold to build up in memory and deserialise them once `continuation` is false. After all, there's no fundamental reason why we can't deserialise the long lines, it's just not what the current implementation does. However, that would allow plugins to make clients consume arbitrary amounts of memory by printing arbitrary amounts of data to stderr, which seems like an undesirable property to introduce into a relatively stable and widely used library.

Instead, this PR allows clients to opt into behaviour that increases memory consumption, and still keeps a hard cap on it at the increased level. It's more fiddly because now clients have to be aware of the maximum log size they expect from a plugin, but also safer so I think the correct decision. Does that reasoning track with you @luoxiaohei, or am I missing an opportunity to simplify here?

luoxiaohei:
@tomhjp Yes, your reasoning tracks with me. I think the configurable option is a good balance between flexibility and safety.

client_test.go
```diff
@@ -1461,18 +1461,13 @@ func testClient_logger(t *testing.T, proto string) {
...
-		stdErrBufferSize = orig
-	}()
-
 	stderr := bytes.Buffer{}
 	c := NewClient(&ClientConfig{
 		Stderr: &stderr,
 		Cmd: &exec.Cmd{
 			Path: "test",
 		},
+		LogBufferSize: 32,
```
tomhjp, line 1470: This tests that the existing behaviour still works, but I can't find any test coverage to confirm that JSON logs are correctly deserialised iff they are within the buffer size. It would be nice to have a test that writes one shorter and one longer JSON log, and see that the level is correctly interpreted for only the shorter one, or something along those lines.

tomhjp (commented):
Thanks! I just re-read the PR description and saw you already discussed my previous question 🤦‍♂️

Looks good, but one naming nit and I'd like to make sure we very explicitly cover the intent of the change in unit tests.

client_test.go
luoxiaohei: Thanks for your suggestion. I will update the Option name and add relevant test cases later.

luoxiaohei:
@tomhjp I've refining the naming and enhancing the unit tests to explicitly cover the intent of the changes. Please take another look when you have a moment. Thanks!

tomhjp (approved):
LGTM, thanks!

Sec32fun32 (approved):
Sec32fun32

PR #267, opened 2023-07-31
Preserve log keys order

oshirohugo:
This pull request aims to preserve the log keys' order provided by the plugin.

Currently, the host unmarshalls the logs emitted by the plugin into a map and uses its keys to create an output that is passed to the host logger. However, the order of the map keys is non-deterministic, resulting in a non-deterministic keys order passed to the host logger.

To address this issue, this pull request extracts the keys in an ordered manner and adds unit tests to ensure that the produced output preserves the order.

### DIFF
--- a/log_entry.go
+++ b/log_entry.go
@@ -5,6 +5,7 @@ package plugin
 
 import (
 	"encoding/json"
+	"regexp"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


oshirohugo:
@fairclothjm would you take a look at this one at you availability? :D

sonamtenzin2:
Hi @oshirohugo, I don't think we should sort the key map just to achieve correct positional ordering in a log. We want to keep the impact of logging as minimal as possible, and sorting would increase the time spent on the same. This could add a considerable time overhead once it accumulates across all logs.

PR #271, opened 2023-08-22
Update docs for creating managed clients

Acconut:
As far as I know, there is no method `NewManagedClient`, but managed clients are created using `NewClient` with `ClientConfig.Managed = true`


fairclothjm (approved):
Thanks!

PR #272, opened 2023-08-24
Make config.Cmd and config.RunnerFunc mutually exclusive

tomhjp:
A set of small follow-ups to #270:

* Make `ClientConfig.Cmd` and `ClientConfig.RunnerFunc` mutually exclusive
* Clients setting `RunnerFunc` can call `ReattachConfig()` but need to supply their own `ReattachFunc` as it can't necessarily be derived from `RunnerFunc`. Exercise `ReattachConfig()` in tests in a way that previously panicked.
* Improve 1 logger line for non-cmd implementations.
* Add `ID()` function to client; accommodates creating a `ReattachFunc` and also useful for client's debug log information.

### DIFF
--- a/client.go
+++ b/client.go
@@ -118,6 +118,19 @@ func (c *Client) NegotiatedVersion() int {
 	return c.negotiatedVersion
 }
 
+// ID returns a unique ID for the running plugin. By default this is the process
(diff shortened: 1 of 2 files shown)


swenson (approved):
LGTM

tomhjp:
Thanks!

PR #273, opened 2023-08-25
Add context argument to runner methods that do work

tomhjp:
* Allows us to properly respect the start timeout for implementations that do long-running work in the Start method.
* Adding context.Context to Wait and Kill makes them more consistent with Start, but is also a bit of a hedge, allowing room in the interface for future features without having to break the API.


swenson (approved):
LGTM

tomhjp:
Thanks!

PR #274, opened 2023-08-28
Revert "Add context argument to runner methods that do work"

jefferai:
Reverts hashicorp/go-plugin#273

This changed a public API without an associated bump in the API version, violating go module rules.

### DIFF
--- a/client.go
+++ b/client.go
@@ -515,7 +515,7 @@ func (c *Client) Kill() {
 
 	// If graceful exiting failed, just kill it
 	c.logger.Warn("plugin failed to exit gracefully")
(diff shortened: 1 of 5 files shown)


jefferai:
Never mind, I didn't notice that runner was also brought in via unreleased code until I saw some older notifications.

In the future it might be good to be more explicit about that in the commit message, otherwise people simply seeing an API change may not understand whether it's to already-released code or not.

tomhjp:
👍 that's correct - I've tried to take my time tagging because of the backwards compatibility commitment I'll be making. I made a note about that for my last intended change before tagging in #275

PR #275, opened 2023-08-28
Add Diagnose to runner.Runner interface

tomhjp:
**Note: This would be a breaking API change, but the `runner.Runner` interface has not been tagged yet. I intend to tag the library after landing this PR, at which point any further changes in the way we use the interface will need to take backwards compatibility into account.**

After working with `runner.Runner` for a bit from main, I've realised the error reporting doesn't make any sense when the plugin fails to start because it's tightly tied to the execution method. In addition, there are quite a few bits of different debug info that would be useful for the container runner from go-secure-stdlib to report, so it seems like a generally useful addition to the Runner interface. Finally, it also cleans up one of the two remaining usages of `ClientConfig.Cmd` that does not take the runner implementation into account.

The last remaining place that `ClientConfig.Cmd` is used without awareness of the runner implementation is in `ReattachConfig()`. However, it's part of the API there because `Pid` is a publicly accessible member of the returned struct, so I'm not planning make further changes there just now.


fairclothjm (approved):

fairclothjm (approved):
👍

PR #277, opened 2023-09-04
Consistently apply Unix socket settings

tomhjp:
Previously, we only supported setting the group for the server-side socket. This change makes it possible to set it on the client side as well. Also fixes a bug where the gRPC broker on the server side would previously not consume the directory/group environment variables.

I found this bug while dropping capabilities for the container runner in the `github.com/hashicorp/go-secure-stdlib/plugincontainer` package, and saw that it was inadvertently relying on the `DAC_OVERRIDE` capability for the plugin to be able to write to client-side sockets it didn't explicitly have permission to.

### DIFF
--- a/client.go
+++ b/client.go
@@ -101,7 +101,7 @@ type Client struct {
 	// forcefully killed.
 	processKilled bool
 
-	hostSocketDir string
+	unixSocketCfg UnixSocketConfig
 }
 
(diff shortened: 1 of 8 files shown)


swenson (approved):
LGTM; great catch!

tomhjp:
Thanks!

PR #281, opened 2023-09-20
Allow more mtls curve

luoxiaohei:
Due to the limitation of `python-grpcio`, which only supports the P256 elliptic curve(https://github.com/grpc/grpc/issues/23235), when AutoMTLS is enabled, the Python protocol plugin service cannot communicate properly with the Go Client.

Therefore, I have added the 'AutoMTLSCurve' option to allow specifying the curve type independently for AutoMTLS in the Client and Server. This will facilitate future extensions.


luoxiaohei:
The limitation of python-grpcio only supporting the P256 curve, since fixed by grpc/grpc#34867, means that the Python protocol plugin service can communicate properly with the Go Client using AutoMTLS as of grpc's next release version, without specifying the curve type. Therefore, this pull request may not be as urgent or necessary as before, while it still provides some flexibility for future extensions. 

Please let me know if you have any questions or feedback.

client.go
```diff
@@ -393,6 +399,10 @@ func NewClient(config *ClientConfig) (c *Client) {
 		})
 	}
 
+	if config.AutoMTLSCurve == nil {
+		config.AutoMTLSCurve = defaultMTLSCurve
```
sonamtenzin2: Instead of changing config in client and server, you can add a default scenario in mtls itself

PR #282, opened 2023-09-21
Add TempDir option to UnixSocketConfig

tomhjp:
When running plugins in containers with systemd's [PrivateTmp=true](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#PrivateTmp=) setting, the containers are not sub-processes so they are not part of the plugin client's file system namespace - as such `/tmp` is different for the 2 sides of the connection and they can't establish communication.

As a workaround, TMPDIR can be set to something that _is_ the same on both sides like `/home/user/tmp`, but that punctures the PrivateTmp setting for all other uses of `/tmp` as well. This lets us choose a single exception.

### DIFF
--- a/client.go
+++ b/client.go
@@ -257,9 +257,10 @@ type UnixSocketConfig struct {
 	// not set, defaults to the directory chosen by os.MkdirTemp.
 	TempDir string
 
-	// The directory to create Unix sockets in. Internally managed by go-plugin
-	// and deleted when the plugin is killed.
-	directory string
+	// The directory to create Unix sockets in. Internally created and managed
+	// by go-plugin and deleted when the plugin is killed. Will be created
+	// inside TempDir if specified.
+	socketDir string
 }
 
 // ReattachConfig is used to configure a client to reattach to an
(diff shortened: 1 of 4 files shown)


swenson (approved):
LGTM.

The title of this PR confuses me a little though. `TempDir` is already a field in `UnixSocketConfig`? It's now we use the option (instead of ignoring, as it appears that it was unused before this change).

tomhjp:
That's a fair comment, the field names were pretty confusing with this addition. `TempDir` is a new field, and we already tracked `directory` (now renamed to `socketDir` to try to improve disambiguation). `TempDir` is the dir that `socketDir` will be created in. `socketDir` is the dir that unix sockets will be created in. We have two layers because we want to both be able to:
1. Specify a base directory that is visible to both the host and plugin
2. And use a unique directory per-plugin so that plugins' unix sockets can be made not visible to each other.

I tried to improve the comments and field names a bit in bf6a95e, but lmk if I didn't succeed.

fairclothjm (approved):

tomhjp:
Thanks!

PR #288, opened 2023-10-16
Support muxing gRPC broker connections over a single net.Conn

tomhjp:
Implements multiplexing for the gRPC broker so that additional gRPC connections are multiplexed onto the plugin's initial listener instead of creating a new Unix or TCP listener for each gRPC stream ID. The primary motivation is the simplification this gives containerised plugins which will only have to negotiate creating a single Unix socket that can be used by both sides, instead of one on each side - in particular this will greatly ease supporting rootless container engines running containers with a non-root user.

This will require both sides of the connection to update their go-plugin version, and as such is strictly opt-in via client config.

The implementation uses the same yamux.Session-based multiplexing that netrpc uses. However, given that we don't have control over when the gRPC library might redial a new connection, we introduce the concept of knocking (essentially a handshake, somewhat similar to TCP's SYN-ACK handshake) to synchronise the client and server on establishing a connection for a pre-agreed stream ID.

I briefly explored implementing gRPC's `grpc.ClientConnInterface` instead to give us full control over when a new connection is established, which would get rid of the need for knocking/pre-agreeing an ID, but I got put off by having to implement `grpc.CallOption`s, and also having to change the library's API, because `GRPCBroker.Dial` currently returns a concrete `*grpc.ClientConn` struct. However, I'm still open to exploring that further if people think those problems are preferable to the ones this PR currently solves.


fairclothjm (approved):
LGTM! I pulled this down and tested with a non-containerized plugin

client.go
```diff
@@ -878,6 +902,20 @@ func (c *Client) Start() (addr net.Addr, err error) {
...
+		if c.config.GRPCBrokerMultiplex && c.protocol == ProtocolGRPC {
+			if len(parts) <= 6 {
+				return nil, errors.New("client requested gRPC broker multiplexing but plugin does not " +
+					"support the feature; for Go plugins, you will need to update the github.com/hashicorp/go-plugin" +
+					"dependency and recompile")
+			}
+			if muxSupported, err := strconv.ParseBool(parts[6]); err != nil {
+				return nil, fmt.Errorf("error parsing %q as a boolean for gRPC broker multiplexing support", parts[6])
+			} else if !muxSupported {
+				return nil, errors.New("client requested gRPC broker multiplexing but plugin does not " +
```
swenson, line 915: Suggestion: put this error as a global variable so that clients can more easily gracefully fallback to setting mux to `false`.

internal/grpcmux/blocked_client_listener.go
```diff
@@ -0,0 +1,48 @@
+package grpcmux
```
swenson, line 1: Amusingly enough, this package doesn't seem to reference or use GRPC. :)

server.go
```diff
@@ -387,6 +388,12 @@ func Serve(opts *ServeConfig) {
 		}
 
 	case ProtocolGRPC:
+		var muxer *grpcmux.GRPCServerMuxer
+		if multiplex := os.Getenv(envMultiplexGRPC); multiplex == "true" || multiplex == "1" {
```
swenson, line 392:
```suggestion
		if multiplex, _ := strconv.ParseBool(os.Getenv(envMultiplexGRPC)); multiplex {
```
?

server.go
```diff
@@ -414,13 +422,27 @@ func Serve(opts *ServeConfig) {
...
+
+		// Old clients will error with new plugins if we blindly append the
+		// seventh segment for gRPC broker multiplexing support, because old
+		// client code uses strings.SplitN(line, "|", 6), which means a seventh
+		// segment will get appended to the sixth segment as "sixthpart|true".
+		//
+		// If the environment variable is set, we assume the client is new enough
+		// to handle a seventh segment, as it should now use
+		// strings.Split(line, "|") and always handle each segment individually.
+		if multiplex := os.Getenv(envMultiplexGRPC); multiplex != "" {
```
swenson, line 442:
```suggestion
		if os.Getenv(envMultiplexGRPC) != "" {
```

swenson (approved):
LGTM, a few minor suggestions

server.go
tomhjp: Nice, yeah that's much cleaner

client.go
tomhjp: Good idea - I used it here and above as well, and also added a test to check we receive the expected error.

swenson (approved):
LGTM!

tomhjp:
Thanks!

Issue #298, opened 2024-02-06
error encountered while scanning stdout: file already closed 


gabivlj:
Similar to https://github.com/hashicorp/go-plugin/issues/116

Started encountering these since 1.5.1 added the error check of the stdout scanner: https://github.com/hashicorp/go-plugin/blob/v1.5.1/client.go#L763

The simple fix would be waiting for this goroutine to end before calling .Wait().

PR #299, opened 2024-02-06
Make sure we wait for stdout/stderr pipes to be closed before calling Wait()

gabivlj:
Fixes https://github.com/hashicorp/go-plugin/issues/298

### DIFF
--- a/client.go
+++ b/client.go
@@ -104,9 +104,9 @@ type Client struct {
 	// goroutines.
 	clientWaitGroup sync.WaitGroup
 
-	// stderrWaitGroup is used to prevent the command's Wait() function from
-	// being called before we've finished reading from the stderr pipe.
-	stderrWaitGroup sync.WaitGroup
+	// pipesWaitGroup is used to prevent the command's Wait() function from
+	// being called before we've finished reading from the stdout and stderr pipe.
+	pipesWaitGroup sync.WaitGroup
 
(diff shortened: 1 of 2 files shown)


gabivlj:
Hi @tomhjp! Could you take a look at these changes whenever you have a chance? Thank you!!

tomhjp:
Thanks for the PR and the ping! I'd like to get a test for this fix, please could you share details on a repro? I can reliably repro if similarly to the previous issue I add a `time.Sleep(time.Second)` inside the stdout scanner loop, but I want to be sure that that's a faithful representation of the real scenario, i.e. is this a race condition that only pops up once in a while, or does it happen every time for your plugin?

tomhjp:
I pushed a test that fails before this fix and passes after in https://github.com/hashicorp/go-plugin/commit/36c021db037a8f9873cca5d914b134fee84440d4. LMK if you think that looks like a good test case.

gabivlj:
Thank you @tomhjp for the test! I appreciate it a lot.
I think it's a good test case.

We were encountering these logs "sometimes" in our error reports so it was just noise.

tomhjp (approved):
Awesome, thanks for the confirmation! In that case I'm convinced we've understood the bug sufficiently well to merge. I'll add the test in a follow up PR. Thanks again for the fix!

PR #300, opened 2024-02-12
Add test for stdout scanner race with runner.Wait()

tomhjp:
Adds a test-case for #299 that reliably fails before the patch, and reliably passes after it.

I ran `go test -count=100 -run ^TestClient_noStdoutScannerRace$ github.com/hashicorp/go-plugin` on commits 586d14f3dcef1eb42bfb7da4c7af102ec6638668 (but with this test added) and 9b348465c6b69c0e402a226c795b1b23a76798af, and got 100/100 failures and 0/100 failures respectively.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -28,10 +28,12 @@ import (
 
 func TestClient(t *testing.T) {
 	process := helperProcess("mock")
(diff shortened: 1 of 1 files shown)


swenson (approved):
LGTM

Love adding a good test.

PR #312, opened 2024-07-04
Remove go-testing-interface from go-plugin

VioletHynes:
I'm a little unclear why the go.sum is like that, I was having issues with `go mod tidy` and it told me to do `go mod tidy -go=1.16 && go mod tidy -go=1.17`, and I did, and now `go mod tidy` seems to work.

Anyway, this removes go-testing-interface from this project. The behaviour should be exactly the same in this project with the Go stdlib version.

For more discussion as to why that's a good thing, see: https://github.com/hashicorp/vault-enterprise/pull/6101 and https://go.hashi.co/rfc/vlt-323. I noticed it was here by it still being an indirect dependency for Vault and I was already in the removal headspace, so I thought why not fix it here too :)

When we get the next version of this in Vault, it will also remove it as an indirect dependency.


fairclothjm (approved):
LGTM

The tests all pass except for TestGRPC_Reflection but that seems to be failing on main as well.

PR #321, opened 2024-10-18
Fix test and add CI

schmichael:
Stole test.yaml from yamux: https://github.com/hashicorp/yamux/blob/49eecf38b432cfb1b543d41c782ea17cca090708/.github/workflows/test.yaml

Not much but better than nothing?


tgross (approved):
LGTM :+1:

PR #323, opened 2024-11-08
feat(client): set client buffer size to avoid buffer overflow

GoldenSheep402:
In client.go

The origin code is:
```go
	go func() {
		defer c.clientWaitGroup.Done()
		defer c.pipesWaitGroup.Done()
		defer close(linesCh)

		scanner := bufio.NewScanner(runner.Stdout())

		for scanner.Scan() {
			linesCh <- scanner.Text()
		}
		if scanner.Err() != nil {
			c.logger.Error("error encountered while scanning stdout", "error", scanner.Err())
		}
	}()
```

If `runner.Stdout()`is bigger than 64kb, it will cause `[ERROR] plugin: error encountered while scanning stdout: error="bufio.Scanner: token too long"`

A fixed version is:
```go
	go func() {
		defer c.clientWaitGroup.Done()
		defer c.pipesWaitGroup.Done()
		defer close(linesCh)

		scanner := bufio.NewScanner(runner.Stdout())
		buf := make([]byte, c.config.ClientScanerBufferSize)
		scanner.Buffer(buf, c.config.ClientScanerBufferSize)

		for scanner.Scan() {
			linesCh <- scanner.Text()
		}
		if scanner.Err() != nil {
			c.logger.Error("error encountered while scanning stdout", "error", scanner.Err())
		}
	}()
```

The `ClientScanerBufferSize` is provided as a config item(default is 64KB), user can set it by themselves if their command will return a large output and avoid receiving a null result.

### DIFF
--- a/client.go
+++ b/client.go
@@ -227,6 +227,10 @@ type ClientConfig struct {
 	// If this is 0, then the default of 64KB is used.
 	PluginLogBufferSize int
 
+	// ClientScanerBufferSize is the buffer size(bytes) to read from stderr for plugin log lines.
+	// If this is 0, then the default of 64KB is used.
+	ClientScanerBufferSize int
+
 	// AutoMTLS has the client and server automatically negotiate mTLS for
 	// transport authentication. This ensures that only the original client will
 	// be allowed to connect to the server, and all other connections will be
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -227,6 +227,10 @@ type ClientConfig struct {
 	// If this is 0, then the default of 64KB is used.
 	PluginLogBufferSize int
 
+	// ClientScanerBufferSize is the buffer size(bytes) to read from stderr for plugin log lines.
+	// If this is 0, then the default of 64KB is used.
+	ClientScanerBufferSize int
```
sonamtenzin2: typo for Scanner

sonamtenzin2 (commented):
Please add test cases for the failure scenario

PR #343, opened 2025-06-10
IND-3869 enabling dependabot

KaushikiAnand:
As per the memo: https://go.hashi.co/memo/eng-004,
Enabling the dependabot for identifying and updating the security and version updates in the repositories rather handling it manually.


ritikrajdev (approved):
LGTM

PR #354, opened 2025-07-18
Generate test protobuf using the `buf` tool

jbardin:
It wasn't clear which versions of the protobuf tooling needed to be used to generate the test code, and updating them required some additional changes. Since the `buf` tool was adopted to generate the example code, use the same tool to make generating the test code easier too.


jbardin:
@hashicorp/team-ip-compliance, waiting on review please

sonamtenzin2 (approved):

PR #387, opened 2026-06-25
ci: Add copywrite workflow and makefile target and fix headers.

jrasell:
Jira: https://hashicorp.atlassian.net/browse/NMD-1551


.github/workflows/copywrite.yaml
```diff
@@ -0,0 +1,24 @@
+name: Check Copywrite Headers
```
thaJeztah, line 1: 😅 `s/CopyWrite/Copyright/` (also in the commit messages)

thaJeztah: Oh! Neverrmind; looks like copywrite is the name for that workflow? Thought it was an accidental slip-up

jrasell: The naming matches the tool name on purpose here: https://github.com/hashicorp/copywrite

tgross (approved):
LGTM

.github/workflows/copywrite.yaml
thaJeztah: Thanks! Yes, I was slow to realise that! Sorry for the noise 🤗

PR #406, opened 2026-09-14
fix(grpc): Ping should check health response status

### LINKED ISSUE #341: GRPCClient Ping method does not check health response
I've been working on a Node based plugin and was testing to see if my health client is working, and I noticed that the current GRPCClient implementation for `Ping()` doesn't actually evaluate if the service is healthy (`SERVING`):

https://github.com/hashicorp/go-plugin/blob/cfdf485783602a2ca85502dedebf441be7bcbc8d/grpc_client.go#L127-L134

Instead of just checking for `err` (which would tell you if the service was "hard down", or didn't implement `/grpc.health.v1.Health/Check`), this should also check the response from `client.Check` to check that it returned `grpc_health_v1.HealthCheckResponse_SERVING`.

Changing this would probably break a lot of people's stuff right now, but I thought I'd point it out.

AruneshDwivedi:
Fixes #341

The GRPCClient.Ping method only checks for transport errors when calling the gRPC health check, but ignores the response status. This means a plugin returning NOT_SERVING would appear healthy from the client perspective.

Changes:
- Capture the HealthCheckResponse in Ping()
- Verify Status == SERVING before returning nil
- Add test confirming NOT_SERVING triggers Ping() error

Note: There is already an open PR (#377) by another contributor with a similar fix. This PR provides the same functional improvement with a regression test.


AruneshDwivedi:
Closing in favor of #377 which addresses the same issue. Thanks @alliasgher!

